PR #3, opened 2020-11-25
Add CI job for checking catalog format

sonicaj:


.github/workflows/format_validation.yml
```diff
@@ -0,0 +1,17 @@
+name: format_validation
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    container:
+      image: sonicaj/catalog_validation:latest
```
william-gr, line 10: Can we deploy it within truenas namespace?

william-gr (approved):

PR #10, opened 2020-12-06
Add Plex chart

sonicaj:


william-gr:
@kmoore134 you probably know more than plex than most of us, so thought ask you for review instead

test/plex/2.1.0/test_values.yaml
```diff
@@ -0,0 +1,72 @@
+# Default values
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# The Image to use for PLEX
+
+image:
+  repository: plexinc/pms-docker
+  tag: 1.20.2.3402-0fec14d92
```
kmoore134, line 9: Do we always have to define a tag here, or is it possible / desirable to instead just pull the latest?

kmoore134 (approved):
Only concern was about hard-coding the plex version and if its possible to just always have it pointed at the 'latest' tag or not. They update pretty frequently.

test/plex/2.1.0/test_values.yaml
sonicaj: We can set `latest` as the default. For this right now, i am using the tag upstream chart is using. However if the user wants to use latest, that's totally fine and is definitely editable by the user.

sonicaj: `questions.yaml` controls if the value can be edited or not and that is from where the defaults are constructed for chart release creations.

kmoore134: Ok, that helps me understand then. In that case it should be OK and user can control at what rate they want to get updates.

PR #14, opened 2020-12-11
Take backup of postgres on helm upgrade

sonicaj:


test/nextcloud/1.1.0/templates/backup-postgres-hook.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - name: POSTGRES_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: db-details
+              key: db-password
+        - name: BACKUP_NAME
+          value: {{ template "postgres.backupName" . }}
+        volumeMounts:
+          - name: postgres-data
+            mountPath: /var/lib/postgresql/data
```
yocalebo, line 37: It's been a minute since I've messed with docker containers, but this is the host path for which the data is actually stored, right? Not inside the container?

sonicaj: Nope, this is the path inside the container.

test/nextcloud/1.1.0/templates/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
       initContainers:
         - name: init-postgresdb
           image: busybox:latest
-          command: ['sh', '-c', "until nslookup {{ template "nextcloud.fullname" . }}-postgres; do echo waiting for postgres; sleep 2; done"]
+          command: ['sh', '-c', "until nc -w 5 -vz {{ template "nextcloud.fullname" . }}-postgres 5432; do echo waiting for postgres; sleep 2; done"]
```
yocalebo, line 42: There is a `_ 5432` in this command? Is that supposed to be there? Looks weird.

sonicaj: Yes it is, we check for 2 things here:
1) Is postgres service available which would be DNS resolving
2) Is postgres listening on 5432 or not

test/nextcloud/1.1.0/templates/backup-postgres-hook.yaml
yocalebo: Okay thanks, do we have a method for maintaining the copies of the databases? i.e . are we only keeping last 2 upgrades worth of backups or is that already being managed with this commit?

test/nextcloud/1.1.0/templates/deployment.yaml
yocalebo: got it, the variable for the container name is stored in between the `{{ }}`. For whatever reason I associated the `5432` to the name of the container.

test/nextcloud/1.1.0/templates/backup-postgres-hook.yaml
sonicaj: So in other words would we end up with a long list of postgres backups and if we are automatically trimming it down ?

yocalebo: Correct, are we managing the copies of the backups that we take.

sonicaj: For now we are not. Backup is only going to happen on chart upgrade which i anticipate is not going to be very often.
But we can make keeping the backup optional when we do restore. So down the road i intend to restore the backup after upgrade of chart completes and then if user wants to store backup of database, we can keep it there, otherwise we can delete the backup file.

test/nextcloud/1.1.0/templates/deployment.yaml
yocalebo, line 42: Apparently this won't work based on @themylogin 's comment on the previous PR that you just merged.

sonicaj: It would work because k8s will restart the container if it fails but still @themylogin is correct and it's better to use postgres container in this case.

test/nextcloud/1.1.0/templates/backup-postgres-hook.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  template:
+    metadata:
+      name: "pre-upgrade-hook2"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: "{{ template "nextcloud.serviceAccountName" . }}"
+      containers:
+      - name: {{ .Chart.Name }}-postgres-backup
+        image: "postgres:13.1"
```
themylogin, line 20: It is repeating itself now, and when doing postgres image upgrade we might change only one and end up with two different postgres images. Do we have an ability to define it as a variable somewhere and use it like `{{ .Chart.Name }}`?

yocalebo (approved):

william-gr (approved):

PR #15, opened 2020-12-11
Use nc instead of nslookup while waiting for postgres to come up

sonicaj:
Using nslookup, the wait time for nextcloud pod to come up can be up to 1-5 minutes. This is reduced with nc to 5-10 seconds.

### DIFF
--- a/test/nextcloud/1.0.0/templates/deployment.yaml
+++ b/test/nextcloud/1.0.0/templates/deployment.yaml
@@ -39,7 +39,7 @@ spec:
       initContainers:
         - name: init-postgresdb
           image: busybox:latest
-          command: ['sh', '-c', "until nslookup {{ template "nextcloud.fullname" . }}-postgres; do echo waiting for postgres; sleep 2; done"]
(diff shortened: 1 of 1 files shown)


yocalebo (approved):

test/nextcloud/1.0.0/templates/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
       initContainers:
         - name: init-postgresdb
           image: busybox:latest
-          command: ['sh', '-c', "until nslookup {{ template "nextcloud.fullname" . }}-postgres; do echo waiting for postgres; sleep 2; done"]
+          command: ['sh', '-c', "until nc -w 5 -vz {{ template "nextcloud.fullname" . }}-postgres 5432; do echo waiting for postgres; sleep 2; done"]
```
themylogin, line 42: With this method it might fail with `The database system is starting up`. You need to either use a real postgres client or just make a container restartable.

sonicaj: @themylogin k8s will restart the container for us if it fails but i agree that it's probably best if i use postgres container here and use `pg_isready` instead. I'll make the required change. Thank you for pointing it out.

PR #26, opened 2021-01-17
Deployments only support "Always" Restart Policy

sonicaj:
https://github.com/kubernetes/kubernetes/issues/24725#issuecomment-214513280


themylogin (approved):

william-gr (approved):

Ornias1993 (approved):
Good catch

PR #28, opened 2021-01-23
Add IPFS chart

kmoore134:
@sonicaj - Fixed this up, seems to validate and work properly with dynamic forms now.


sonicaj (commented):
It looks really good, however can we please have this under `test` train ? I'll be updating `charts` train today or tomorrow hopefully as soon as we have the dynamic charts PR finalized, and this can be moved there. Reasoning is that current UI is not able to handle the creation form for this chart and users on 20.12 or master won't be able to install this.

test/ipfs/1.0.0/README.md
```diff
@@ -0,0 +1,22 @@
...
+|:-------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------|
+| `image.repository`                               | Image repository                                                                                                                        | `ipfs/go-ipfs`                    |
+| `image.tag`                                      | IPFS image tag. Possible values listed [here](https://hub.docker.com/r/ipfs/go-ipfs/tags).                                              | `v0.8.0-rc1`   |
+| `image.pullPolicy`                               | Image pull policy                                                                                                                       | `IfNotPresent`                   |
+| `extraArgs`                                      | Additional command line arguments to pass to the IPFS server                                                                            | `[]`                             |
```
sonicaj, line 22: We are not using `extraArgs` ?

sonicaj: Also this is not really required plus it has a maintenance overhead as we would have to keep questions.yaml and this table in sync. We can probably have this updated by CI or something similar down the road and remove it even for now ?

test/ipfs/1.0.0/templates/deployment.yaml
```diff
@@ -0,0 +1,124 @@
+{{ $scheme := "http" }}
```
sonicaj, line 1: Is this required ?

test/ipfs/1.0.0/templates/service.yaml
```diff
@@ -0,0 +1,31 @@
+{{ $scheme := "http" }}
```
sonicaj, line 1: Is this required ?

test/ipfs/1.0.0/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "ipfs.serviceAccountName" . | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    app: {{ template "ipfs.name" . }}
+    chart: {{ template "ipfs.chart" . }}
+    release: "{{ .Release.Name }}"
```
sonicaj, line 9: `{{ .Release.Name | quote }}` should be better in this case I think

sonicaj (commented):
LGTM, just a few minor cosmetic nits

sonicaj (approved):
Actually I think i can clean up the above bits for all the charts we have as we would have these concerns there as well. Thank you :)

kmoore134:
> Actually I think i can clean up the above bits for all the charts we have as we would have these concerns there as well. Thank you :)

Even better, since I copied this from minio chart, and it'll be good to get them all into proper shape at same time.

Thanks @sonicaj !

PR #29, opened 2021-01-28
Update charts stable train

sonicaj:
( We will be merging this on the day of release as UI cannot currently adapt to the new format dynamically - this is a concern for especially ipfs chart )


sonicaj:
Closing this due to the delayed 21.02 release. We can have the newer version of test train in which utilizes library chart.

PR #30, opened 2021-01-28
Introduce Library Chart for Official Catalog

sonicaj:
This PR adds following changes:

1) Introduce library chart to keep charts DRY.
2) Have ports static for ipfs containers and instead rely on service to make it dynamic.
3) Add migrations to bring catalog items to the same format.


Ornias1993:
Why not base it on the great example from k8s-at-home?

sonicaj:
@Ornias1993 most of the charts there only have 1 deployment and are not really complex charts. These charts however have an upstream which is quite complex and specific to the application and we do one day envision to allow most if not all of the features upstream offers and that won't work well in that case

Ornias1993:
@sonicaj Good point indeed, I also can't use that common chart for most of my charts (traefik, mariadb) for the same reason

That being said: Thanks for the headsup, i'll rename my common chart to truecharts-common for clarity :)

Ornias1993:
@sonicaj You might want to checkout 
https://github.com/truecharts/truecharts/blob/dev/library/common/templates/custom/_appStorage.tpl

Info here:
https://github.com/truecharts/truecharts/pull/41

kmoore134 (approved):
LGTM! Much cleaner than before, thanks for doing this!

Ornias1993 (approved):
Codewise it looks good, could be more DRY imho, but not bad :)
However the update dependency script could use a little explaination. what is it for and how can it be used?

sonicaj:
Thanks for reviewing @Ornias1993, i agree it can be more DRY but that also would mean losing leverage on changing things later and then we would just have to rework this anyways.
About the dependency script, it's a helper tool which will update all the dependencies of all catalog items in a train with all their versions, so that dev would not have to do that for each version of a catalog item in a train.

william-gr (approved):

PR #36, opened 2021-02-19
Change versioning for official apps

sonicaj:
This PR introduces following changes:

1) Fixes a chart version to an application version i.e fixed docker tag which is not user configurable.
2) Properly updates app version so that it can be reflected correctly in the UI for chart releases.

Motivation for these changes is that a chart version is not guaranteed to work with every version of docker image available for the particular application and with this change we can have sane versioning for official applications at least.

This also helps control versioning on chart release upgrades i.e if user is using 1.18 image tag and then on upgrade even if we update default value of tag, that is not going to change for the user as schema would consider the value already present as user supplied and override it with default.


Ornias1993 (approved):
This would significantly reduce the amount of support calls, in case users (try to) run charts with incompatible app versions.

Code is solid.

Ornias1993:
Is the processing of ix_values.yaml already integrated in TrueNAS SCALE and catalog_validation?

sonicaj:
@Ornias1993 it is however there is one commit from this PR https://github.com/truenas/middleware/pull/6456 which is also required. Moving on, this expectation can not hold for charts/stable train and I think we can only safely say that most if not every user will have this change until the next TN Scale release.

kmoore134 (approved):

william-gr (approved):

PR #53, opened 2021-05-07
Render ix-chart in the UI dynamically

sonicaj:
This PR adds changes to render ix-chart in the UI dynamically similar to how we handle rest of the applications. This reduces code/complexity of handling ix-chart in the UI. Earlier it was done in a static way  because of time constraints. The changes introduced here specifically hide advanced fields from the UI.


rick-mesta (approved):

Ornias1993 (approved):
Hmm... I think I'm required to approve this, as this was originally my proposal :P

PR #54, opened 2021-05-08
Add official application for Chia

sonicaj:


kmoore134:
Testing on latest nightly images, clicking to enable the host-path volume never shows anything. Not sure if issue with the app, or my nightly version from 5.7.2021 :)

image

sonicaj:
@kmoore134 it's a UI bug, none of the other apps are working as well wrt this. I'll file a jira issue, thank you for pointing it out :) Meanwhile can you please confirm if host path is not used, the app works as desired ?

kmoore134:
> @kmoore134 it's a UI bug, none of the other apps are working as well wrt this. I'll file a jira issue, thank you for pointing it out :) Meanwhile can you please confirm if host path is not used, the app works as desired ?

Makes sense. Went ahead and tested without the host volumes and it seemed to work fine, great work!

kmoore134 (approved):

PR #57, opened 2021-05-14
Update tag to 1.1.5

daniel-gallagher:
Updated tag to most recent Chia release. This release contains important bug fixes.

### DIFF
--- a/charts/chia/1.0.0/test_values.yaml
+++ b/charts/chia/1.0.0/test_values.yaml
@@ -1,6 +1,6 @@
 image:
   repository: ghcr.io/chia-network/chia
-  tag: 1.1.4
+  tag: 1.1.5
   pullPolicy: IfNotPresent
 updateStrategy: Recreate
 environmentVariables:


daniel-gallagher:
I was moving too fast and meant to just open this PR on my fork for testing, so please feel free to disregard... Though the `1.1.5` release is a pretty important upgrade, as the Chia team has stated using any version less might result in blocks being rejected by peers.

> We were not checking for negative values in the uint64 constructor. Therefore coins with negative values were added to the mempool. These blocks passed validation, but they did not get added into the blockchain due to negative values not serializing in uint64. Farmers making these blocks would make blocks that did not make it into or advance the chain, so the blockchain slowed down starting at block 255518 around 6:35AM PDT. The fix adds a check in the mempool and block validation, and does not disconnect peers who send these invalid blocks (any peer 1.1.4 or older), making this update not mandatory but is recommended. Users not updating might see their blocks get rejected from other peers.

REF: https://github.com/Chia-Network/chia-blockchain/releases/tag/1.1.5

daniel-gallagher:
Just going to close this until I actually have a PR worth submitting.

Issue #56, opened 2021-05-11
[Feature] Add an extra plotting host path for Chia


jfturcot:
Plotting on ZFS is much faster than on other filesystems on regular spinning hard drives, but I would like to be able to use a separate SSD only pool for plotting. Could you please add an extra host path for that use case?

Thank you for creating this Chia app, it works a lot better than their Docker container and was a nice surprise to see it in the official applications after fighting with their container for a few days.

sonicaj:
@jfturcot to be clear, you would like to be able to specify a host path for `plot` volume ? `plot` volume is what's mounted inside the pod/container at `/plots`

jfturcot:
Sure that could work, or maybe /plots/tmp since they call it a tmp_dir on the command line. Thanks!

sonicaj:
So the application already supports it - we have a UI bug where that option is not showed when you click on the boolean value which says use host path volume. You can use the command line to test it out

```
midclt call -job chart.release.create '{"item": "chia", "values": {"appVolumeMounts": {"plots": {"hostPathEnabled": true, "hostPath": "/mnt/evo/chia/plots"}, "config": {"hostPathEnabled": true, "hostPath": "/mnt/evo/chia/config"}}}, "release_name": "chia", "catalog": "OFFICIAL"}'
```

In this case both `config` and `plot` volumes are consuming host paths.

sonicaj:
Please let me know if you run into issues with the above command.

PR #59, opened 2021-05-17
Allow specifying host path volumes/environment variables for chia

sonicaj:
Closes https://github.com/truenas/charts/issues/56


test/chia/1.1.0/Chart.yaml
```diff
@@ -14,5 +14,5 @@ sources:
 - https://github.com/orgs/chia-network/packages/container/package/chia
 dependencies:
   - name: common
-    repository: file://../../../library/common/2105.0.0
-    version: 2105.0.0
```
Simoneu01, line 18: why downgrade common version?

sonicaj: It's test train, so this change does not affect it. However i've still bumped it. Thanks for pointing it out!

rick-mesta (approved):

PR #71, opened 2021-06-14
Add metadata to allow automated updates of catalog item(s)

sonicaj:


rick-mesta (approved):
From my limited understanding, this looks 👍🏼 ... but I'd feel a warm fuzzy if I had @yocalebo's eyes on it as well.

yocalebo (approved):

rick-mesta:
Thx for the ⚡ fast review, @yocalebo 🥰

Ornias1993:
Just to be clear: This looks to be updates of catalog item Image Tag, not the actual catalog items themselves, right?
Aka: It updates the tags, but does not automatically update the catalog item when newer versions of the catalog item are available, right?

sonicaj:
@Ornias1993 this basically helps automate pushing newer versions of catalog item(s) when a newer image tag is available for a catalog item. `catalog_update` can be used which uses these metadata files and then makes a PR with the newer catalog item(s) using newer image tags ( if available ).

PR #79, opened 2021-07-04
Update test train catalog items

sonicaj:
This PR adds changes to update catalog items in test train to reflect newer tags already present in charts/stable train. Moving on, we should ideally be making all the changes in test train and then copying them as it is to stable train to keep the trains in sync and also make sure there are no regressions for stable train.


rick-mesta (approved):
This LGTM... I'd feel more comfy with a second pair of eyes from any other mw folk out there 🙏🏼

PR #81, opened 2021-07-06
Add HTTPS support for nextcloud

sonicaj:


Ornias1993:
Wouldn't this time be much beter spend supporting ingress? :-P

PR #82, opened 2021-07-07
Update chia to the TruePool.io / iXsystems sponsored image which

kmoore134:
includes pooled plot support in pre-release.


sonicaj (commented):
Can we please update `test_values.yaml` as well so CI can run using the updated image tag ?

kmoore134:
> Can we please update `test_values.yaml` as well so CI can run using the updated image tag ?

Done, does that look correct now?

PR #117, opened 2021-08-10
NAS-111810 / 21.08 / Fixes #116 by swapping MINIO_ACCESS_KEY and MINIO_SECRET_KEY

### LINKED ISSUE #116: MINIO_ACCESS_KEY and MINIO_SECRET_KEY are deprecated.
```
2021-08-10 18:13:07.001116+00:00WARNING: MINIO_ACCESS_KEY and MINIO_SECRET_KEY are deprecated.
2021-08-10 18:13:07.001330+00:00Please use MINIO_ROOT_USER and MINIO_ROOT_PASSWORD
```

joshuacox:
with MINIO_ROOT_USER and MINIO_ROOT_PASSWORD


joshuacox:
@sonicaj still sort of a WIP, I had to update the default image tag, and I think there may be other context differences from the official catalog that I need to correct.

charts/minio/1.3.11/default_values.yaml
```diff
@@ -2,7 +2,7 @@
 ##
 image:
   repository: minio/minio
-  tag: RELEASE.2020-11-19T23-48-16Z
+  tag: RELEASE.2021-08-05T22-01-19Z
```
sonicaj, line 5: This file is not used, let's remove this please

charts/minio/1.3.11/questions.yaml
```diff
@@ -72,21 +72,21 @@ questions:
           schema:
             type: string
 
-  - variable: accessKey
-    label: "Access Key"
+  - variable: rootUser
```
sonicaj, line 75: We can't change the variable name as otherwise how would existing users access codes still work after upgrade to the newer version of minio app ?
If we want to change the variable name still, we should be writing a migration then ( please see migration directory ). If you want to take that direction, please let me know i can clarify more.

charts/minio/1.3.11/questions.yaml
```diff
@@ -137,7 +137,7 @@ questions:
             type: int
             min: 9000
             max: 65535
-            default: 9002
+            default: 9001
```
sonicaj, line 140: This conflicts with nextcloud app from official catalog, let's please keep `9002` as the default value

charts/minio/1.3.11/templates/deployment.yaml
```diff
@@ -46,8 +46,8 @@ spec:
             {{ $envList = mustAppend $envList (dict "name" "MINIO_BROWSER_REDIRECT_URL" "value" (printf "%s://%s:%d" (include "minio.scheme" .) .Values.minioDomain (.Values.service.consolePort | int))) }}
             {{ $envList = mustAppend $envList (dict "name" "MINIO_SERVER_URL" "value" (printf "%s://%s:%d" (include "minio.scheme" .) .Values.minioDomain (.Values.service.nodePort | int))) }}
             {{ end }}
-            {{ $envList = mustAppend $envList (dict "name" "MINIO_ACCESS_KEY" "valueFromSecret" true "secretName" $secretName "secretKey" "accesskey") }}
-            {{ $envList = mustAppend $envList (dict "name" "MINIO_SECRET_KEY" "valueFromSecret" true "secretName" $secretName "secretKey" "secretkey") }}
+            {{ $envList = mustAppend $envList (dict "name" "MINIO_ROOT_USER" "valueFromSecret" true "secretName" $secretName "secretKey" "rootUser") }}
```
sonicaj, line 49: Have you updated the secret where the actual access codes are stored ? As this key won't exist in the secret at all

sonicaj (changes requested):

joshuacox:
@sonicaj ok, this was probably too ambitious, so I am reducing to merely changing out the environment variables which will silence the error message.  And I also changed out the wording of the questions that the users sees, to match what they will see on the login page to the minio ui (username and password not access and secret key), as that might lead to confusion.  Everything else is untouched.

sonicaj (commented):
These look good, however can you please bump the catalog item version as well ?
So basically rename 1.3.11 to 1.3.12 and also please update Chart.yaml to reflect the newer version.

charts/minio/1.3.11/templates/deployment.yaml
joshuacox: I changed this back to the way it was and I am reducing to merely changing out the environment variables which will silence the error message. And I also changed out the wording of the questions that the users sees, to match what they will see on the login page to the minio ui (username and password not access and secret key), as that might lead to confusion.

joshuacox:
> These look good, however can you please bump the catalog item version as well ?
> So basically rename 1.3.11 to 1.3.12 and also please update Chart.yaml to reflect the newer version.

done deal move to 1.3.12

sonicaj (changes requested):
Can you please rename the minio app version under test train as well ?

sonicaj (approved):
Thanks!

PR #137, opened 2021-09-04
add optional resource limits to Machinaris Chart on `test` track

g-a-c:
I was having trouble with memory usage on my low-powered NAS, so this allows me to specify a CPU limit and memory limit (in Kubernetes syntax, e.g. `500m` for 500 millicores and `1Gi` for 1 Gibibyte of RAM), and I thought this might be useful for others so committed it to the `test` track initially


test/machinaris/1.0.6/questions.yaml
```diff
@@ -157,3 +159,20 @@ questions:
                 label: "Value"
                 schema:
                   type: string
+  - variable: enableResourceLimits
+    label: "Enable Pod resource limits?"
```
sonicaj, line 163: Let's please remove the question mark

test/machinaris/1.0.6/questions.yaml
```diff
@@ -157,3 +159,20 @@ questions:
...
+    group: "Resource Limits"
+    schema:
+      type: boolean
+  - variable: cpuLimit
+    label: "CPU limit"
+    group: "Resource Limits"
+    schema:
+      type: string
```
sonicaj, line 171: Do you think we should perhaps validate if specified value is valid using regex https://github.com/truenas/charts/blob/master/docs/schema/string.md ?

sonicaj (commented):
Thank you for your contribution, just a minor nit

test/machinaris/1.0.6/questions.yaml
g-a-c: Both values should now be regex-validated, the memory value comes straight from within the Kubernetes/`k3s` codebase, the CPU value is mine but has been tested with [Regex101](https://regex101.com/r/0PQNCH/1) to make sure it passes and fails on expected values

sonicaj (approved):
Thanks!

PR #138, opened 2021-09-06
NAS-112140 / 21.10 / Update nextcloud upgrade strategy

sonicaj:
For nextcloud tags, 21.1 is greater then 21.1.0 so we reverse sort the tags list before doing semantic versioning as in semantic versioning both tags are equal and when we reverse sort it, 21.1 is prioritised over 21.1.0.


themylogin:
How cool is that stuff like this can be configured per-app, once more congratulations on a great software design!

themylogin (approved):

PR #147, opened 2021-09-14
Add Collabora Application

sonicaj:


Ornias1993 (commented):
Interesting build, am I correct in assuming the NGINX container is mostly just for TLS support?

PR #277, opened 2021-12-28
NAS-114011 / 22.02 / Allow using plexpass for plex app

sonicaj:


charts/plex/1.6.16/ix_values.yaml
```diff
@@ -2,3 +2,7 @@ image:
   pullPolicy: IfNotPresent
   repository: plexinc/pms-docker
   tag: 1.25.2.5319-c43dc0277
+plexpass_image:
```
Qubad786, line 5: shouldn't this be camelCase?

Qubad786 (approved):

charts/plex/1.6.16/ix_values.yaml
sonicaj: Yes, that's a better practice with charts - i'll make the change. Thanks for pointing it out

Qubad786 (approved):

PR #334, opened 2022-02-13
Fix incorrect port and change version to 1.1.5

andrew-stclair:
Corrects gateway port mismatch between the port configured in TrueNAS interface and the port the gateway runs on.

Gateway now runs on port 9880 instead of 9080. Interface still uses port 9880 by default


andrew-stclair:
Fixed merge conflicts but have not updated version number

sonicaj (commented):
Thank you @andrew-stclair for your contribution, can you update your branch from master please and make the change ? Thanks!

PR #389, opened 2022-03-24
NAS-114424 / 22.12 / Add netdata application

Qubad786:


sonicaj (commented):
LGTM, however can we please add upgrade strategy for this ? Thanks!

sonicaj (approved):
Thanks!

PR #416, opened 2022-04-12
NAS-115758 / 22.12 / Remove immutable on storage field for netdata

Qubad786:
## Problem

Right now UI is allowing `immutable` fields to be modified and hence sending payload to `chart.release.update` which results in the app not being able to be modified.

## Solution

For now, removing `immutable` flag on netdata storage section and making a UI ticket to address this so that we can consume this again allowing netdata app to be modified.


sonicaj (approved):
I don't like it very much but i am not seeing a workaround as well otherwise for the time being :(

PR #428, opened 2022-04-19
NAS-115858 / 22.12 / Fix netdata hitting apparmor profile

miguelangel-nubla:
Netdata hitting apparmor is spamming logs:
```
Apr 19 21:52:04 truenas kernel: audit: type=1400 audit(1650397924.087:7477502): apparmor="DENIED" operation="ptrace" profile="docker-default" pid=2185738 comm="apps.plugin" requested_mask="read" denied_mask="read" peer="unconfined"
```

[Official netdata chart](https://github.com/netdata/helmchart/blob/master/charts/netdata/templates/daemonset.yaml) uses unconfined apparmor declaration.


Qubad786 (commented):
@miguelangel-nubla your contribution is much appreciated!
Can you make the following changes as well:
1. Update `test` train and bump the app version to `1.0.4` and update `Chart.yaml` accordingly to reflect the newer version
2. Bump the app version in `charts` train and corresponding `Chart.yaml` change as well

Ornias1993:
I wonder, in our experience SCALE does not actually use AppArmour, or is this a nightly things @Qubad786 ?

Ornias1993:
@Qubad786 We've also added netdata and also have no issues like this... Afaik apparmour is also not enabled, so I do not see the need for this.

sonicaj:
ping @Qubad786

Qubad786:
@miguelangel-nubla can you email me a debug of your system when you see this issue at mrehan@ixsystems.com please so we can investigate/see the exact configuration case which is causing this?

I was not able to reproduce this either..

miguelangel-nubla:
I updated my system, currently in TrueNAS-SCALE-22.02.2.1 and can no longer reproduce the issue.
I originally reported it while running SCALE-22.12-BETA.1, so must have been fixed along the way.

Ornias1993:
> I originally reported it while running SCALE-22.12-BETA.1, so must have been fixed along the way.

I think you're mistaken there. As that version does not even exist, because 22.12 isn't technically even in ALPHA.1 yet.

PR #441, opened 2022-04-25
NAS-115807 / 22.12 / Add syncthing application

Qubad786:


sonicaj (approved):

anodos325:
Out of curiosity have we done much testing of syncthing with our product regarding features we have but other OSes might not?

For example NFSv4 ACLs and large xattrs?

anodos325:
Or is this sort of quasi-supported on our side?

anodos325:
> It works, but it writes it's data as it's own user (if correctly configured). Which often does not work out well with shares.
It's somewhat worse that that from a quick glance at the docs. It also does not sync relevant file metadata (EAs / xattrs) which may result in unusable files in some cases (think MacOS with resource forks). ACLS (NFSv4 and POSIX1e are stripped as well). So very significant limitations imho.

Ornias1993:
> > It works, but it writes it's data as it's own user (if correctly configured). Which often does not work out well with shares.
> 
> It's somewhat worse that that from a quick glance at the docs. It also does not sync relevant file metadata (EAs / xattrs) which may result in unusable files in some cases (think MacOS with resource forks). ACLS (NFSv4 and POSIX1e are stripped as well). So very significant limitations imho.

Agreed, the target audience is syncing basic files for a single(!) user, that's an important limitation many of our syncthing users also encountered.

It's not a bad application at all, other sync solutions are often much worse, but before adding Apps to the Official catalog it's worth while thoroughly reviewing it's interactions within the ecosystem. As users would come to iX for help if they encounter anything they find "weird"... imho ofcoarse.

PR #479, opened 2022-05-31
NAS-116291 / 22.12 / Allow adding custom host path for postgres for nextcloud

Qubad786:


charts/nextcloud/1.6.1/questions.yaml
```diff
@@ -237,17 +236,33 @@ questions:
...
                   default: "ix-postgres_data"
+                  show_if: [["hostPathEnabled", "=", false]]
                   editable: false
+
               - variable: mountPath
                 label: "Postgresql Data Mount Path"
                 description: "Path where the volume will be mounted inside the pod"
                 schema:
                   type: path
-                  editable: false
```
sonicaj, line 249: We should add `hidden` attr but not remove this one as we don't want the user to modify this

sonicaj (approved):

PR #519, opened 2022-06-15
Add host path validations for applications

Qubad786:
## Context

Added validation for all apps so that they use paths which are considered safe. Following validation has been added:

1. Validate paths are not consuming locked dataset
2. Ensuring host path mounts come under `/mnt/tank/` and root dataset paths are also not allowed


sonicaj (approved):

sonicaj:
@kmoore134 yes, we also should allow /dev and /sys mounts. Will have that change ready soonish :)

PR #523, opened 2022-06-16
Fix typo

darkfeline:


sonicaj (changes requested):
Thank you @darkfeline for your contribution. You would need to bump the common chart version and after doing that, update all the charts as well so that they consume the newer common chart.

darkfeline:
Are there any tools or guides for doing that?  seems like a lot of manual work for a typo fix

PR #542, opened 2022-06-20
NAS-116797 / 22.12 / Allow special paths for host path mounts

Qubad786:
## Context

We should allow special paths like `/sys` or `/dev` or `/cluster` as being valid paths which can be mounted inside applications.


yocalebo (changes requested):
How are we preventing applications from accessing the ctdb shared volume directory inside /cluster?

Qubad786:
Can you clarify which path inside the cluster path should be restricted and i can add the changes?

Qubad786:
Also does it mean that top level `/cluster` should not be allowed to mount as we don't want to allow mounting a subpath?
I am not sure how `/cluster` looks and if we disable some specific path to be mounted inside the apps, what other apps do we want to allow as clearly it means top level `/cluster` should not be allowed as well..

yocalebo:
In middlewared repo I have a global variable "CTDB_LOCAL_MOUNT" that needs to be excluded from being mounted by any app.

Also, I'm not sure adding /cluster right now is smart. We are making too many changes that could potentially be backwards incompatible and I don't really want to think about 3rd party apps right now.

Qubad786:
So in that case should we not allow adding `/cluster` then @yocalebo @sonicaj?

About the local mount, i assume any path except for `/cluster/ctdb_shared_vol` is safe inside `/cluster`?

yocalebo:
> So in that case should we not allow adding `/cluster` then @yocalebo @sonicaj?
> 
> 
> 
> About the local mount, i assume any path except for `/cluster/ctdb_shared_vol` is safe inside `/cluster`?

Yes, I don't think we should add /cluster right now. Also, I don't want to hardcode the ctdb shared volume fuse mountpoint in this repo. We need to be able to update that singular variable and have everywhere else get updated as expected. But adding /cluster is premature imo and we should hold off until we understand/design/test a little more.

library/common/2112.0.0/templates/lib/storage/_hostpathValidation.tpl
```diff
@@ -2,19 +2,25 @@
 Validate list of host path in a specific format
 */}}
 {{- define "common.storage.hostPathsValidation" -}}
+    {{- $allowed_paths := (list "mnt" "sys" "dev" "cluster") -}}
     {{- range . -}}
         {{- $host_p := splitList "/" . -}}
         {{- $host_p := (without $host_p "") -}}
         {{- $error_msg := (printf "Invalid hostpath %s. Path must be a valid path under a given pool e.g `/mnt/tank/somepath` is valid whereas `/mnt` or `/mnt/tank` are invalid examples." .) -}}
-        {{- if lt ($host_p | len) 3 -}}
+        {{- if and (eq (index $host_p 0) "mnt") (lt ($host_p | len) 3) -}}
```
yocalebo, line 10: Is it correct syntax to do `if and`?

yocalebo (approved):
I don't like this but I guess we don't have an option 😄

PR #554, opened 2022-06-25
Create jellyfin

hzykj:


sonicaj:
@hzykj please go through the following link to see how we accept new applications in the official catalog:
https://github.com/truenas/charts/blob/master/contributing.md

If there are any queries, let me know. Thanks!

PR #562, opened 2022-06-27
NAS-116635 / 22.12 / Add storj application

Qubad786:
## Context

It was requested that we add `storj` application to official catalog. The app now uses the following workflow:

1. On installation of the application, a job will be executed which will generate storj [identity](https://docs.storj.io/node/dependencies/identity/).
2. Deployment of `storj` will remain pending until identity generation has been completed.
3. Once identity has been generated, `storj` will deploy with `apps` user being the user on the filesystem and the process owner.

How-to on deploying application https://docs.google.com/document/d/1oa2P1CwqYPN6L_uDUYjvt-0JGDG7T6d45rm2MMmn0Wc/edit?usp=sharing


Ornias1993:
1. This would run at each restart, once. Not only at install. You can, however, use an initcontainer
2. If you eventually add probes (which is a best-practice) this might lead to issues.

---

Instead it might be best to run this as a initcontainer, limited to install-only and, if you want to scale this to multiple containers, add a lockfile to prevent double-execution.

Qubad786:
@Ornias1993 it would not run on each restart as restart in SCALE means scaling deployments/statefulsets up/down.

About doing this in init-container, we basically want to run this as `apps` user but the identity generation pod as `root` user as the binary does not allow us to change the path of where the signed certs would be located and it uses current user's home directory for that and we don't want to maintain a custom image.

sonicaj:
@Qubad786 after discussing with @yocalebo we have decided to hold off on this feature and make it bluefin only. What we would like to do is the following:

1. Allow setting memory limit size on the volumes while defaulting to something small
2. Have a maximum limit on the size which can be set for these volumes and we can discuss the maximum with the team as to what it should be ( it would probably require middleware changes )

sonicaj (approved):
I can confirm this works nicely, good work @Qubad786!

One minor nit, i think we better be explicit on the init container for identity creation that it needs to run as root. Can you please make that change ? Thanks!

Darkknight33:
Putting this here for visibility of the original PR author: https://github.com/truenas/charts/issues/762

PR #575, opened 2022-07-03
NAS-115275 / 22.12 / Allow users to specify their own webui portals for apps 

sonicaj:


test/ix-chart/2207.1.0/questions.yaml
```diff
@@ -23,9 +23,64 @@ groups:
...
+    description: "Enable webui portal for easier access to workload ( Only valid for TrueNAS SCALE Bluefin )"
+    group: "Portal Configuration"
+    schema:
+      type: boolean
+      default: false
+  - variable: portalDetails
+    label: "WebUI Portal"
+    description: "Configure WebUI Portal"
+    group: "Portal Configuration"
+    schema:
```
Ornias1993, line 41: might be best to put a dict under this, for easy multi-portal support in the future.
This extra dict layer doesn't have to be used yet, but gives a lot more flexibility.

Ornias1993 (changes requested):

Qubad786 (approved):

Issue #505, opened 2022-06-10
Plex 1.7.7 hostpath validation breaks installs with external data mounts


anotb:
The addition of hostpath validations break installs where a hostpath created in /mnt from an external share mounted using smb/nfs, rclone or mergerfs was being used as a data library. 

Specifically in charts/plex/1.7.7/questions.yaml, the addition of 
                        $ref:
                          - "validations/hostPath"

Any chance that can not be made mandatory? Would hate to have to change hostpath and have plex rescan the library. 

**Error log -** 

Error: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 412, in run
    await self.future
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 448, in __run_body
    rv = await self.method(*([self] + args))
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 1261, in nf
    return await func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 1129, in nf
    res = await f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/chart_releases_linux/upgrade.py", line 116, in upgrade
    await self.upgrade_chart_release(job, release, options)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/chart_releases_linux/upgrade.py", line 266, in upgrade_chart_release
    config, context = await self.middleware.call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1308, in call
    return await self._call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1257, in _call
    return await methodobj(*prepared_call.args)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/chart_releases_linux/chart_release.py", line 332, in normalise_and_validate_values
    dict_obj = await self.middleware.call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1308, in call
    return await self._call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1257, in _call
    return await methodobj(*prepared_call.args)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/chart_releases_linux/validation.py", line 69, in validate_values
    verrors.check()
  File "/usr/lib/python3/dist-packages/middlewared/service_exception.py", line 62, in check
    raise self
[...]

PR #646, opened 2022-08-03
NAS-117484 / 22.12 / Allowing memory backed volumes in Plex

Qubad786:
## Context

It was requested to allow memory backed volumes similar to what `/dev/shm` does.


sonicaj (approved):

william-gr:
What is the intent of this?

sonicaj:
@william-gr we had a request for this in https://github.com/truenas/charts/issues/505

william-gr:
We usually don't do things only because users requested them...
Allowing memory to be arbitrarily used may not fly very well and cause all sorts of weirdness, without important constraints.

We are getting to a point we should at least discuss a feature with a supervisor/manager before implementing it, as much as it may seem trivial.

I will defer to @yocalebo on this one

sonicaj:
@Qubad786 after discussing with @yocalebo we have decided to hold off on this feature and make it bluefin only. What we would like to do is the following:

1. Allow setting memory limit size on the volumes while defaulting to something small
2. Have a maximum limit on the size which can be set for these volumes and we can discuss the maximum with the team as to what it should be ( it would probably require middleware changes )

Ornias1993:
> @Qubad786 after discussing with @yocalebo we have decided to hold off on this feature and make it bluefin only. What we would like to do is the following:
> 
> 1. Allow setting memory limit size on the volumes while defaulting to something small
> 2. Have a maximum limit on the size which can be set for these volumes and we can discuss the maximum with the team as to what it should be ( it would probably require middleware changes )

Please note that, while some validation is great, it's also important to not-break the behavior users expect from Kubernetes and docker.

We've two options:
1. The DIY option (aka: pick your own path)
How we somewhat solved the issue @william-gr hinted at (TLDR: Users fucking things up): is that we have the feature basically hidden away by having it as a subtype of the "emptyDir" type.
Simply put, users first have to make something an emptyDir and then also have to manually put it on a ramdisk. In our experience this leads to most users doing this either 
A. Had to be very intenting to do precisely what they where doing.
B. Knew what they where doing before hand.

Making it "too easy" like this PR, for an App that would be **insanely** ram hungry, like plex... is super tricky...

2. `/dev/shm`
Most linux-savy Users expect this to be a ramdisk, so we made it a ramdisk by default.

PR #647, opened 2022-08-04
NAS-117324 / 22.12 / [DRAFT]: Make apps to run as nonroot

Qubad786:
### Problem
Some apps are running were creating`mnts` with `root` ownerships, that can create problems.

### Solution
Making sure apps are running as `non root` user wherever possible.

#### Steps taken
1. Running pod containers as `non root` via security context (using `app` user).
2. Few app defined variables are used (to make sure `app`) where apps were running specific flows starting with `root` user.
3. Wrote migration to `chown` for previously created volumes.

PS: For fresh installations `chown` is done too.


charts/netdata/1.0.10/migrations/migrate
```diff
@@ -0,0 +1,117 @@
...
+
+def chown_path(uid, gid, host_path):
+    acltool = subprocess.run([
+        "/usr/bin/nfs4xdr_winacl",
+        "-a", "chown",
+        "-O", str(uid), "-G", str(gid),
+        "-r",
+        "-c", host_path,
+        "-p", host_path], check=False, capture_output=True
+    )
```
Ornias1993, line 84: Question:
Does this support: Non-ACL, POSIX-ACL *and* NFSv4 ACL's?

Ornias1993: @sonicaj Maybe you can related, does this support all three possible permissions variants?

Ornias1993 (commented):
While the code in essence looks good, I want to highlight a few notes:
- Having migration copies for validation/permissions setting everywhere is quite inefficient
- It might be prudent to move to PVC storage as a default storage instead. As this offloads the whole permissions shebang to kubernetes using `fsgroup` which works like a dream.
- The other Apps, that use LSIO containers, use plain chmod internally, which would not (always) play nice with ACL's. So you basically get a seperating in how permissions are handled for the Official Apps, which might not be fully clear for the end-user. For some of the problematic containers there are decent alternatives available though.

charts/home-assistant/1.0.25/migrations/migrate
```diff
@@ -0,0 +1,117 @@
...
+            dataset_names.append(data["datasetName"])
+    for ix_volume in values["ixVolumes"]:
+        validate_path(values["ownerUID"], values["ownerGID"], ix_volume["hostPath"], values["release_name"])
+        if any([volume_name in ix_volume["hostPath"] for volume_name in dataset_names]):
+            host_paths.append(ix_volume["hostPath"])
+    return host_paths
+
+
+def migrate(values):
+    for host_path in validate_host_volume_path(values):
```
Ornias1993, line 106: For big folders this is gonna wreak havoc on execution times though.

charts/netdata/1.0.10/migrations/migrate
Ornias1993: Verified to be the case.
Resolved.

Ornias1993:
@Qubad786 I'm assuming you plan to port this to the new permissions API?

PR #648, opened 2022-08-04
NAS-117445 / 22.12 / Make minio data mount path configurable

Qubad786:
### Description
For users who want to run minio in distributed mode may want to configure data directory url. For this reason we make minio data directory's mount path user configurable.


sonicaj (approved):
LGTM, however please update your branch before merging as master has moved forward. Thanks!

PR #672, opened 2022-08-11
NAS-117625 / 22.12 / Add "run as user" option in ix chart

Qubad786:
### Description
- Run as user option in `ix-charts`.


charts/ix-chart/2208.0.0/templates/_workload.tpl
```diff
@@ -60,6 +60,8 @@ containers:
   stdin: {{ .Values.stdin }}
   securityContext:
     privileged: {{ .Values.securityContext.privileged }}
+    runAsUser: {{ .Values.securityContext.runAsUser }}
```
sonicaj, line 63: Let's please make this optional i.e if user has not specified any overrides, we should not specify a default which we are doing right now for root and let the image uid/gid take preference. Thanks!

sonicaj (approved):

PR #697, opened 2022-08-24
[home-assistant] dnsPolicy and hostNetwork

voidlily:
When `hostNetwork` is set to true, `dnsPolicy` should be set to `ClusterFirstWithHostNet` explicitly.

### DIFF
--- a/charts/home-assistant/1.0.29/templates/deployment.yaml
+++ b/charts/home-assistant/1.0.29/templates/deployment.yaml
@@ -30,7 +30,10 @@ spec:
         {{- include "common.labels.selectorLabels" . | nindent 8 }}
       annotations: {{ include "common.annotations" . | nindent 8 }}
     spec:
+    {{- if .Values.hostNetwork }}
       hostNetwork: {{ .Values.hostNetwork }}
+      dnsPolicy: ClusterFirstWithHostNet
+    {{- end }}
       initContainers:
         - name: init-postgresdb
           image: {{ template "postgres.imageName" . }}


charts/home-assistant/1.0.29/templates/deployment.yaml
```diff
@@ -30,7 +30,10 @@ spec:
         {{- include "common.labels.selectorLabels" . | nindent 8 }}
       annotations: {{ include "common.annotations" . | nindent 8 }}
     spec:
+    {{- if .Values.hostNetwork }}
```
sonicaj, line 33: Let's please move this to after `hostNetwork:` as once it's set and it's unset again - i don't think it will turn it off by default and the older value would carry on unless explicit `hostNetwork: false` is specified

Ornias1993:
@sonicaj It might be more efficient to handle both in your common chart, as the configuration format of these is always the same on each and every application.

charts/home-assistant/1.0.29/templates/deployment.yaml
stavros-k: Not super clean, but for now will address the issue.

We can discuss to include this in common.
```yaml
{{- if .Values.hostNetwork }}
      hostNetwork: true
      dnsPolicy: ClusterFirstWithHostNet
{{- else }}
      hostNetwork: false
{{- end }}
```

Ornias1993: dnsPolicy should also be configurable outside of hostNetwork.
Maybe check copy the compatible 5 lines from TC Common for now?

stavros-k:
Closing this, as it's solved in #1081 
Thanks

PR #708, opened 2022-08-30
Update machinaris to 0.8.3

Qubad786:


meilon:
Shouldn't charts/machinaris/1.1.12/ix_values.yaml not also Update the app version?

PR #717, opened 2022-09-04
Fix typo in qbittorrent

jab416171:


sonicaj:
This will be handled in https://github.com/truenas/charts/pull/718. Thanks!

PR #770, opened 2022-10-03
NAS-118428 / 22.12 / Add upgrade strategy for storj app

Qubad786:
## Context

Upgrade strategy allows automatic upgrades of applications.


sonicaj (commented):
Let's please be consistent in quotes usage and either use single quotes or double but not both together. Thanks!

sonicaj (approved):
LGTM, however seeing how complex this logic can get - probably better if we allow specifying unit tests for upgrade strategies. I'll create a ticket for that. Thanks!

PR #823, opened 2022-11-03
NAS-118859 / Add LogSearchApi to Minio App

stavros-k:
This adds the logsearch api functinality.
There was previously a container just for that, but MinIO merged it with the operator container.

Used that container and override the entrypoint to start just the logsearchapi binary.

Flow goes like this:
- MinIO posts on logsearchapi's webhook audit logs.
- logsearchapi stores those audit logs on pgsql
- MinIO uses the query URL to fetch those audit logs and display them under `Monitoring` -> `Audit Logs`

Enabling this functionality requires a postgres database.
- I'm spinning one only when logsearchAPI is enabled.
- Switch pgsql to recreate instead of rolling update, so we don't have 2 active pgsql instances, as that can corrupt the database.

- Regarding UI, user has this option 
image

Disk Capacity only shown when logsearch is checked.
Also has this for postgres data:
image

I'm not sure how we  can hide those when logsearchapi is not enabled.


stavros-k:
Tagged you both for review, for a second opinion, as I'm not yet 100% familiar with the common library.

test/minio/1.7.0/ix_values.yaml
```diff
@@ -0,0 +1,13 @@
+image:
+  pullPolicy: IfNotPresent
+  repository: minio/minio
+  tag: RELEASE.2022-10-24T18-35-07Z
```
Qubad786, line 4: Version needs to be changed now

test/minio/1.7.0/questions.yaml
```diff
@@ -184,6 +184,28 @@ questions:
       $ref:
         - "definitions/certificate"
 
+  - variable: logsearchapi
+    label: Log Search API Configuration
+    description: Log Search API Configuration
+    group: Minio Configuration
+    schema:
+      type: dict
+      additional_attrs: true
```
Qubad786, line 193: We should avoid using this as it means user can add additional attrs which will never be validated and will remain there always.
I suppose you used it because you have specified defaults in `ix_values` which we can change in that case and define them here instead?

Qubad786 (commented):
LGTM! Couple of minor nits

test/minio/1.7.0/questions.yaml
stavros-k: It was only added, because I remember having validation issues on TrueCharts with it.
Can you please explain what it does exactly? I couldn't find anything regarding this.

Qubad786: So a dict can only have attributes which are explicitly listed in it but if you specify this, it means that users can specify any other attrs they like `a` or `xyz` or something else and it won't be validated by middleware because it's not defined in the schema..
You would probably be having issues with it because of how you specify defaults in `ix_values.yaml` which means that you should be changing that as it won't be good practice to allow users to specify `a/b/c` etc to a dict.

Ornias1993: > So a dict can only have attributes which are explicitly listed in it but if you specify this, it means that users can specify any other attrs they like `a` or `xyz` or something else and it won't be validated by middleware because it's not defined in the schema.. You would probably be having issues with it because of how you specify defaults in `ix_values.yaml` which means that you should be changing that as it won't be good practice to allow users to specify `a/b/c` etc to a dict.

Which makes it borderline impossible to write clean helm-charts.

You should be open to the standard helm practice:
Defaults stay in values.yaml (in this case ix_values.yaml) and the USER configuration stays in questions.yaml.

However, your statement that users can randomly add dicts is completely untrue and values added in ix_values.yaml will **not** stay "forever" either.

We're doing this for TrueCharts for absolute ages and it works flawlessly for 1.5 year by now (including changing values in ix_values.yaml), so i think i'm qualified to say you're wrong here ;-)

Ornias1993: Note: the second you prevent questions.yaml to require 5 rows to define 1 row of values.yaml defaults, we can talk about using it. But the current design is not usable and is completely non-standard.

I also want to note that this practice is **NOT** compatible with the common chart that is designed as agreed with @sonicaj

stavros-k: > So a dict can only have attributes which are explicitly listed in it but if you specify this, it means that users can specify any other attrs they like `a` or `xyz` or something else and it won't be validated by middleware because it's not defined in the schema.. You would probably be having issues with it because of how you specify defaults in `ix_values.yaml` which means that you should be changing that as it won't be good practice to allow users to specify `a/b/c` etc to a dict.

Okay, I see, spoke with Waqar too and got few more questions solved too.

test/minio/1.7.0/ix_values.yaml
```diff
@@ -2,12 +2,12 @@ image:
   pullPolicy: IfNotPresent
   repository: minio/minio
   tag: RELEASE.2022-11-10T18-20-21Z
+logSearchImage:
```
Ornias1993, line 5: note: this is required anyway for renovate to grab this in the future... ;-)

test/minio/1.7.0/questions.yaml
```diff
@@ -252,6 +273,85 @@ questions:
                   type: hostpath
                   required: true
 
+  - variable: postgresAppVolumeMounts
```
Qubad786, line 276: Should user be allowed to change host path for postgres once it has deployed already?

Qubad786: Basically any change to underlying storage for an already initialized postgres?

stavros-k: > Should user be allowed to change host path for postgres once it has deployed already?

No, not really. 

But if we remove this completely, users can't define a hostPath at all. It will always be stored in ix-volumes.

Other option is to write extra helm code, that only takes into account the path only when it's at an `isInstall` phase.
But then it gets messy real quick. You have to save the path defined the first time into a secret/configmap and retrieve it on each redeploy/upgrade. And also you will now have a path field that does nothing beyond the initial install.

Unless there is a way to lock (editable: false) the path field after the initial install.

sonicaj: @stavros-k iirc the schema attr of interest in that case is `immutable` - that should cater to this use case. Also i think there are lots of other places in other apps which can benefit from the same change. So we can create a ticket to cater to those as well.

stavros-k: Oh nice! 

Added `immutable` only on the postgres data path.
Not in the postgres backup path.

Backup path can be changed, without issues.

Qubad786 (approved):

PR #845, opened 2022-11-07
NAS-112453 / Adds WG-Easy App (Server)

stavros-k:
Currently only 1 instance is able to run and work.
A second instance with changed ports will appear active, and it's webUI will be accessible.

Only the **last** instance that starts will be working normally, 
any previous instance will _stop working or not working correctly_.

This is because the interface created by this app (`wg0`) has it's name hardcoded.

There was a request for that to be configurable upstream but until today, there isn't anything like that implemented.
https://github.com/WeeJeWel/wg-easy/issues/222


stavros-k:
There are 2 `sysctl` properties that need to be set.
- `net.ipv4.conf.all.src_valid_mark=1`
- `net.ipv4.ip_forward=1`

The second appears to be set in my system, first is not.

We could either document how the user can add those or have an initContainer set them.

test/wg-easy/1.0.0/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+      app: {{ template "common.names.name" . }}
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      name: {{ template "common.names.fullname" . }}
+      labels:
+        {{- include "common.labels.selectorLabels" . | nindent 8 }}
+      annotations: {{ include "common.annotations" . | nindent 8 }}
+    spec:
+      hostNetwork: {{ .Values.hostNetwork }}
```
sonicaj, line 28: With `hostNetwork` enabled, i suppose we would have different routes/rules being added to host network's namespace - we should be cleaning it up when the app is scaled down. Thoughts ?

test/wg-easy/1.0.0/templates/delete-route-post-delete-hook.yaml
```diff
@@ -0,0 +1,33 @@
+{{ $values := (. | mustDeepCopy) }}
+{{ $_ := set $values "common" (dict "nameSuffix" "wg-easy") }}
+{{ $ip := .Values.wgeasy.client_address_range | replace "x" "0" }}
+apiVersion: batch/v1
+kind: Job
```
sonicaj, line 5: Let's please probably use the following hook instead
```
spec:
  containers:
    - name: containerName
      lifecycle:
        preStop:
          exec:
           command: [
```

test/wg-easy/1.0.0/templates/delete-route-post-delete-hook.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: post-delete-routes-wg-easy
+  annotations:
+    "helm.sh/hook": post-delete
```
sonicaj, line 9: This would only run when the app is deleted, if the app is scaled down we don't want leftover routes by wireguard lying around.

test/wg-easy/1.0.0/questions.yaml
```diff
@@ -0,0 +1,259 @@
...
+    description: Set CPU/memory limits for Kubernetes Pod
+
+portals:
+  web_portal:
+    protocols:
+      - http
+    host:
+      - $node_ip
+    ports:
+      - $variable-web_port
```
sonicaj, line 20: Admin portal will be broken i assume as `web_port` is not a valid variable in schema

test/wg-easy/upgrade_strategy
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/python3
+import json
+import sys
+import re
+
+from catalog_update.upgrade_strategy import datetime_versioning
+
+
+# Upstream uses only a single digit for tags
+VERSION_REG = r'[0-9]*'
```
sonicaj, line 10: I assume this has been copied over ;) The upgrade strategy will fail to get latest tags - i think we can remove this for now and add it later separately in a subsequent PR.

sonicaj (commented):
Good work, some minor nits!

test/wg-easy/upgrade_strategy
stavros-k: I've copied an update_strategy script and modified the regex to match single digit for tags (thats what wg-easy uses)
If you think it won't work, yea I can remove it for now!

test/wg-easy/1.0.0/questions.yaml
stavros-k: Fixed!

test/wg-easy/upgrade_strategy
sonicaj: Yup it wont as the newer mapping function below was for a different application. We can go through this later in a subsequent PR

test/wg-easy/1.0.0/templates/delete-route-post-delete-hook.yaml
stavros-k: I've left the post-delete hook also, just to be 100% that on delete it cleans it.
Let me just modify the script a bit

stavros-k: Done

test/wg-easy/1.0.0/templates/delete-route-post-delete-hook.yaml
```diff
@@ -0,0 +1,36 @@
+{{ $values := (. | mustDeepCopy) }}
+{{ $_ := set $values "common" (dict "nameSuffix" "wg-easy") }}
+{{ $ip := .Values.wgeasy.client_address_range | replace "x" "0" }}
+apiVersion: batch/v1
```
sonicaj, line 4: We should be removing this job, this is not required anymore ?

test/wg-easy/1.0.0/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+                    echo "Deleting routes created by the app..."
+                    netmask=$(ip route | grep {{ $ip }})
+                    netmask=$(echo $netmask | grep -o -E '/.\d*')
+                    netmask=${netmask#/}
+                    echo "Matched routes to delete... {{ $ip }}/$netmask"
+                    # Don't try to delete routes if steps above didn't grep-ed anything
+                    if [ ! "$netmask" == "" ]; then
+                      ip route del {{ $ip }}/$netmask || echo "Route deletion failed..."
+                    fi
+                    echo "Routes deleted..."
```
sonicaj, line 107: We should be deleting the wireguard interface as well

sonicaj: Also what would happen if 2 instances of the same app are installed ? Do we have a conflict over interface name ?

test/wg-easy/1.0.0/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+            readOnlyRootFilesystem: false
+            runAsNonRoot: false
+          volumeMounts: {{ include "common.storage.configureAppVolumeMountsInContainer" .Values | nindent 12 }}
+          {{ range $index, $hostPathConfiguration := .Values.extraAppVolumeMounts }}
+            - name: extrappvolume-{{ $index }}
+              mountPath: {{ $hostPathConfiguration.mountPath }}
+          {{ end }}
+          ports:
+            - name: udp
+              containerPort: 51820
```
sonicaj, line 51: Also with host network enabled, hard coding this would not work as users can't change ports because container's network is essentially host's network and it will always be hardcoded to these ports.
We should probably remove the node port service as well in that case as it's redundant because host network is used directly

stavros-k: Missed that, thanks!

Made service to not render at all when hostNet is enabled.

test/wg-easy/1.0.0/templates/delete-route-post-delete-hook.yaml
stavros-k: Had it there as fail safe mostly. But yea it's redundant

test/wg-easy/1.0.0/templates/deployment.yaml
stavros-k: 2 Instances with on hostNet won't work.
As the GUI port is not configurable. and it's port will conlict with the existing instance.


On app removal wg0 is removed, will check now if it also gets removed on scale down

stavros-k: Modified it a bit. App can use different port for both the WG port and webUI.
But the problem is that the `wg0` interface name is hardcoded, and I don't see a way to modify that.
So while technically you can run 2 instances (or more). It's a limitation from the app that it's using a hardcoded interface name
---

Extended the script a bit to also clear the interface on scale down and/or deletion.

Ornias1993 (commented):
"This is because the interface created by this app (wg0) has it's name hardcoded."

Won't this interfere with the wireguard adapter generated for use with TrueCommand Cloud?

sonicaj:
@Ornias1993 it definitely will. We would probably have to change the naming in middleware for truecommand and mark this app for bluefin only.

stavros-k:
> @Ornias1993 it definitely will. We would probably have to change the naming in middleware for truecommand and mark this app for bluefin only.

Good idea, probably prefix it with something like `ix-` to avoid any conflicts, as users might still install other apps from Launch Docker that might use wg* interfaces.

test/wg-easy/1.0.0/questions.yaml
```diff
@@ -0,0 +1,259 @@
...
+                  editable: true
+                  default: /etc/wireguard
+              - variable: hostPathEnabled
+                label: Enable Custom Host Path for WG-Easy Configuration Volume
+                schema:
+                  type: boolean
+                  default: false
+                  show_subquestions_if: true
+                  subquestions:
+                    - variable: hostPath
```
sonicaj, line 201: Let's make this immutable as well please

test/wg-easy/1.0.0/questions.yaml
```diff
@@ -0,0 +1,259 @@
...
+                      label: Host Path for WG-Easy Configuration Volume
+                      schema:
+                        type: hostpath
+                        required: true
```
stavros-k, line 205:
```suggestion
                        required: true
                        immutable: true
```

PR #848, opened 2022-11-08
"/admin/" is a non-existing path for Photoprism.

bobdubois:
"/admin/" is a non-existing path for Photoprism. "/" is the correct path.


Qubad786:
Thank you @bobdubois for the PR - there were some more changes needed..(https://github.com/truenas/charts/pull/849)

PR #864, opened 2022-11-14
NAS-118902 /  Update MinIO Changelog

stavros-k:
MinIO released a Major Feature release, which removes some components that the existing users are using.

Reverting to the last known versions that supports that mode. So existing installations function.

https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html


Ornias1993 (approved):
Agreed as discussed.
Does highlight the need for more CI testing on this repo though...

sonicaj (commented):
What will happen to users who have got the latest version and now we try to upgrade them to an older version ? Does it break/work ?

Ornias1993:
> What will happen to users who have got the latest version and now we try to upgrade them to an older version ? Does it break/work ?

Worst case scenario you would have a user with a broken setup due to the upstream change, also having a broken update due to this change.

stavros-k:
> What will happen to users who have got the latest version and now we try to upgrade them to an older version ? Does it break/work ?

Few comments here suggest that the rollback will work.
https://github.com/minio/minio/discussions/15967

The error mentioned in the ticket, should stop the instance of performing any actions to the data.
So rollbacks should work.

But if it did let the upgrade instance to perform any actions, damage is already done.

Ornias1993:
@sonicaj I've looked at NAS-118902 itself and I think it's pretty clear that current version isn't going to run anyway indeed.
So this change basically is risk-free.

sonicaj (approved):
LGTM!

PR #870, opened 2022-11-16
NAS-119024 / Remove unneeded ports (80/443) from emby and fix hostnetwork

stavros-k:
- Removes 80/443 ports, Emby does not use those at all.
- Hides the port configuration when host net is enabled
- removes `hostPort`, not needed when using `NodePort`
- Adds configmap for the portal, so port is assigned dynamically based on hostnet or not
- When hostnet is enabled, spins services as ClusterIP incase other apps wants to have internal cluster access


Ornias1993 (approved):
LGTM, no issues here except correcting mistakes from the past.

Ornias1993 (approved):

test/emby/1.0.14/templates/service-tcp.yaml
```diff
@@ -1,8 +1,10 @@
 {{ $svc := .Values.embyServerHttp }}
 {{ $ports := list }}
-{{ $ports = mustAppend $ports (dict "name" "emby" "port" $svc.port "nodePort" $svc.port "targetPort" "emby") }}
-{{ $ports = mustAppend $ports (dict "name" "http" "port" 80 "targetPort" "emby") }}
-{{ $ports = mustAppend $ports (dict "name" "https" "port" 443 "targetPort" "emby") }}
+{{ if .Values.hostNetwork }}
```
Qubad786, line 3: Do we need the service at all when using host network?

stavros-k: Only if we want other apps in the cluster accessing emby via ClusterIP.
So it wont have to use the hostnetwork to communicate.

For example, tautulli would communicate with Plex for some stats and info, having an internal channel (ClusterIP) would avoid going through all the network layers to reach it via host net.

I don't know if there are any simillar apps/use cases for emby.

Ornias1993: There is no downside though…

Qubad786: Makes sense

Qubad786 (approved):
LGTM!

Please update your branch from master before merging

PR #921, opened 2022-12-13
Update catalog items ha, plex

stavros-k:


stavros-k:
Updated those 2, so nextcloud can be isolated on update PR. 
To investigate why it's failing

PR #926, opened 2022-12-15
storj: update to v1.68.2

jcassette:
v1.66.4 was a pre-release, so use a proper one.

Signed-off-by: Julien Cassette <julien.cassette@gmail.com>


stavros-k:
Hey, thanks. Will take care of it in #937 as it needs some extra steps.

Thanks

PR #934, opened 2022-12-19
NAS-119472 / 23.10 / Remove `min` for ports in UI, but validate with template

stavros-k:
Currently UI limits port to 9000, but that doesn't apply when using hostNetwork.
And can be misleading, since an app that consumes a port less than 9000, will do so when hostNetwork is enabled.

Removed the minimum port validation from UI and added a validation in the templates, that will only show a message for <9000 port when NOT using hostNetwork


charts/ix-chart/2212.0.1/questions.yaml
```diff
@@ -427,7 +426,6 @@ questions:
                 schema:
                   type: int
                   required: true
-                  min: 9000
```
sonicaj, line 430: This is not required ? 
Node port services are not used with host networking iirc ?

stavros-k: You are correct. It's not even showing when hostNet is enabled.

sonicaj (approved):
Thanks!

PR #936, opened 2022-12-20
NAS-119496 / 23.10 / fix nextcloud secret and update

stavros-k:
Fixed: 
- Secret on nextcloud for `db-details` which was missing the `b64enc` on `db-user`
- Made the random generation of password more robust.
- Postgres Database name defaults to postgres user, while the username (as `nextcloud`) was used in all nextcloud references. the postgres user was set to `postgres`. So postgres was creating a `postgres` database but nextcloud was trying to connect to `nextcloud` database.

Updated:
- nextcloud container
- nginx container

Note: All values under `data` in secrets must be `base64` encrypted.


Qubad786 (commented):
@stavros-k do upgrades to this newer version break older setups?

stavros-k:
> @stavros-k do upgrades to this newer version break older setups?

Correct observation...

I'm confused as to how it worked currently.
This is the env value for the current postgres, also check the postgres password

image

And this is how nextcloud config looks like
image

So yea, it's defintely breaking older setups. Which means we also need to change major version number at least.

Ornias1993:
Though the underlaying problem is the fact that nothing is actually install-tested at the moment.
I would suggest waiting for the new common and merging this + the new common + actuall tests all-at-once.

Ornias1993:
Last change could work @stavros-k :)

stavros-k:
Did the following test after my last change:

- Created a new branch.
- Copied `.14` NC chart 
- Pasted to `.15`
- Applied changes to `.15`
- Removed catalog.json
- Added repo/branch to Scale
- Installed .14
- Logged in the UI created files/user/etc
- Upgraded to .15
- Everything worked fine.
- Also checked env variables and stayed all the same

Ornias1993 (approved):
While far from perfect, this at least solves the issue for now. 
LGTM 👍

Qubad786 (approved):
LGTM 👍

PR #937, opened 2022-12-20
NAS-119497 / 23.10 / storj-update-stable

stavros-k:
As mentioned in #926, `v1.66.4` was a pre-release version.

Updating to a stable version


Qubad786 (approved):
Why didn't upgrade strategy work 🤔?

stavros-k:
> Why didn't upgrade strategy work thinking?

Their releases don't follow a scheme that can differentiate pre-releases from releases, for that reason we disabled the update strategy

Qubad786:
@stavros-k let's re-enable it then?

stavros-k:
> @stavros-k let's re-enable it then?

It's still the case. If we re-enable them we will just get daily prompts from the bot to update to a pre-release

PR #946, opened 2022-12-27
Pull minio changes from `test` to `charts`

stavros-k:


charts/minio/1.7.0/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: '2022-11-10'
+appVersion: '2022-10-24'
```
Qubad786, line 2: Why are we downgrading minio version?

stavros-k: Good catch, copied from the test train which looks like haven't been changed there.
Fixed it here, it will be fixed tomorrow on test train too with the bugclerk updates

Qubad786 (approved):

PR #964, opened 2023-01-19
Still learning:  How do I get back to synced with the "Sonicaj" commits being part of this pull? 

dbb12345:
So I made my normal Machinaris update but have these Update Catalog Information which I'm pretty sure I don't want pushed to the main.  Sorry if a stupid question but how do I/can I, break out the Machinaris update?  Hope that makes sense,

### DIFF
--- a/catalog.json
+++ b/catalog.json
@@ -56,9 +56,9 @@
             "location": "/__w/charts/charts/charts/ipfs",
             "healthy": true,
             "healthy_error": null,
-            "latest_version": "1.2.11",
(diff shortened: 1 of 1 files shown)


stavros-k:
> So I made my normal Machinaris update but have these Update Catalog Information which I'm pretty sure I don't want pushed to the main. Sorry if a stupid question but how do I/can I, break out the Machinaris update? Hope that makes sense,

Hey, thanks for the contribution. You would need to `rebase` your branch based on the master branch of `truenas/charts` repo.
So it only contains your changes.

PR #968, opened 2023-01-21
pihole: Fix DHCP and HostNetwork

nhukc:
There are several bugs that prevent pihole from being used as a DHCP server on the host's network.

- Pihole requires NET_ADMIN to enable DHCP.
- Pihole requires the start, end, and router address to be filled in to enable DHCP.
- When the pihole pod is started in the root networking namespace, port 80 is already taken by the TrueNAS webui.
- When the pihole pod is started in the root networking namespace, it only allow DHCP/DNS traffic over the eth0 interface. 

Fixing these issues allows pihole to function as a DHCP server for the host's network. To test this, enable DHCP server and enter the required fields and check the "Enable Host Network" checkbox.

Screenshot from 2023-01-21 05-41-56

### DIFF
--- a/charts/pihole/1.0.12/questions.yaml
+++ b/charts/pihole/1.0.12/questions.yaml
@@ -65,6 +65,30 @@ questions:
       default: 20722
       required: true
       show_if: [["dhcp", "=", true]]
+  - variable: dhcp_start
+    label: "DHCP Start Address"
+    group: Networking
+    schema:
+      type: ipaddr
+      cidr: false
+      required: true
+      show_if: [["dhcp", "=", true]]
+  - variable: dhcp_end
+    label: "DHCP End Address"
+    group: Networking
+    schema:
+      type: ipaddr
+      cidr: false
+      required: true
+      show_if: [["dhcp", "=", true]]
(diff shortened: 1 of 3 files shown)


garrettgu10:
LGTM!

stavros-k (commented):
Thanks for your contribution. Added some comments, feel free to share your opinion if you see anything wrong.

nhukc:
Everything looks good to me. Thanks.

Nusephived:
Hello,

Thanks for these changes, when it will be available on True NAS ? (last pihole version from 2022-12-01)

Screenshot 2023-02-07 at 18 33 37

Ornias1993:
@stavros-k DHCP (a layer 2 protocol) can only work with hostNetworking enabled. Splitting those off into two seperate settings is kinda weird.

Basically DHCP should be a "show if" of hostNetworking.

Also:
You can just always set the WEB_PORT to something non-root like 8080, as is basically best-practice anyway.

stavros-k:
Fixes have been added in #1049.
Closing this now.
Thanks for your contribution.

PR #977, opened 2023-02-03
NAS-119718 / 23.10 / Adds lifecycle hook to install ffmpeg in NC

stavros-k:
Tested on new installation and ffmpeg got installed successfully.
Also tested to stop/start the app to make sure it does not fail with the package already installed.

Enabled the preview Generator app, 
Added the movie preview provider in the config.
run manually the preview generation and it generated thumbnail for mp4 and webp movie


Ornias1993:
This is a very bad idea, will not work when the app ever is going to be ran rootless, leads to non-reproducable deployments.

It also is in stark contrast to the advice given by Nextcloud themselves, which is an iX Partner.
Is it really the wish of iX to fubar this App and antagonise partners with bad design choices?

Ornias1993 (changes requested):
This is a very bad idea, will not (and cannot be made to) work when the app ever is going to be ran rootless, leads to non-reproducable deployments.

It also is in stark contrast to the advice given by Nextcloud themselves, which is an iX Partner.
Is it really the wish of iX to fubar this App and potentially antagonise a partner with bad design choices?

stavros-k:
Converted to draft, in order to add a "checkbox" on questions to enable this

charts/nextcloud/1.6.17/templates/deployment.yaml
```diff
@@ -82,6 +82,20 @@ spec: {{ include "common.deployment.common_spec" . | nindent 2 }}
         - name: extrappvolume-{{ $index }}
           mountPath: {{ $hostPathConfiguration.mountPath }}
         {{ end }}
+        {{- if .Values.nextcloud.install_ffmpeg }}
+        lifecycle:
+          postStart:
```
Qubad786, line 87: Should we have a probe perhaps to make sure not to mark the app as ready when this is being installed?

Qubad786 (commented):
Only charts train has been bumped, can we make the same changes to test train too?

charts/nextcloud/1.6.17/templates/deployment.yaml
Ornias1993: This runs after probe greenlights it.
Helm does not support adding your own custom states between deploying and ready.

There is a reason this is a bad idea.

Qubad786: @stavros-k can we see if we can wait for this in a probe..
https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks it appears it should be independent of probes 🤔

stavros-k: > Should we have a probe perhaps to make sure not to mark the app as ready when this is being installed?

I've added probes, I only check on the startupProbe for the ffmepg command.
On liveness and readiness I do a check on the status.php page

Qubad786 (approved):

Issue #967, opened 2023-01-21
Machinaris out of date


Tuckie:
Please update to 0.8.6

https://github.com/guydavis/machinaris/blob/main/CHANGELOG.md#086---2023-01-03

Thank you!

stavros-k:
> Greetings! Author of Machinaris here. Any chance this application template could use the[:latest Docker image tag](https://github.com/guydavis/machinaris/pkgs/container/machinaris/61540430?tag=latest), rather than pinning an older version by number? This would remove the hassle of manually changing version numbers in the application template for you. Thanks much!

Hello, I've opened a PR to update the version tag, and also enabled the automatic update strategy.

As a best practice, I'd like to avoid using latest tag, and pin it to a tagged version.

stavros-k:
I'm closing this issue as `Machinaris` will no longer be maintained officially.
Thanks

blllc:
Between you guys and truecharts with all their nonsense, you're really winning hearts and minds. How about you, oh, i dunno... try helping instead of just saying "meh, we're done with it?" You add things, people use them, then you take them away? About as weak as it gets.

It's getting to the point where keeping TrueNAS up-to-date and viable is untenable. I'm sure you don't care, but well, you should. People will eventually migrate away from the platform because of crap like this. It's annoying, counterproductive, and childish. 

Rant over.

PR #983, opened 2023-02-08
Update machinaris container tag

stavros-k:
Updates machinaris to the latest version and re-enables update_strategy

Fixes https://github.com/truenas/charts/issues/967


dbb12345 (changes requested):
Can we up the Machinaris version to v0.8.7 so we are on the latest version before merge?

PR #988, opened 2023-02-09
Updated TrueNAS SCALE Branding

Qouo:
Updated TrueNAS "Scale" to "SCALE" to be consistent with iX Brand Style Guide


Qubad786 (approved):
Thanks @Qouo ✅

PR #999, opened 2023-02-22
NAS-120452 / Add MinIO Enterprise

stavros-k:
> `catalog.json` will be restored


.github/ct-install-config/chart_schema.yaml
```diff
@@ -17,7 +16,8 @@ condition: str(required=False)
 keywords: list(str(), required=False)
 tags: str(required=False)
 maintainers: list(include('maintainer'))
-annotations: map(str(), str(), required=False)
+annotations:
+  title: str(required=True)
```
stavros-k, line 20: https://helm.sh/docs/topics/charts/
```As of [v3.3.2](https://github.com/helm/helm/releases/tag/v3.3.2), additional fields are not allowed. The recommended approach is to add custom metadata in annotations```

PR #1002, opened 2023-02-24
NAS-120452 / 23.10 / Add MinIO Enterprise

stavros-k:
- [x] Support Single Node - Single Drive 
- [x] Support Multi Node - Multi Drive (Distributed Mode)
- [x] Support Single Node - Multi Drive (Same code/logic as the above, just few lines added in the description to differentiate the format)
- [x] Support consuming scale certificate. 
- [x] Run by default as 568 user 
   - [ ] Make it user configurable?
- [x] Have installContainer to chown ixVols or hostPaths to 568 (Limit resources to 1cpu / 512mb ram)
- [x] Support changing the resources (CPU, Memory) for the MinIO container
- [x] Support LogSearchAPI (and Postgres, runs as 999 user, can be any UID, but has some drawbacks on initialization)
- [x] Have installContainer to chown ixVols or hostPaths to 999 for postgres (Limit resources to 1cpu / 512mb ram)
- [x] Have initContainers to wait for postgres and logsearch api to be ready when logsearch is enabled
- [x] Support Automatic Postgres Data Backup on upgrade

---

This PR targes `common-fixes-improvements` branch as I have some fixes there

---

Marked it as ready for review as the app itself is ready for it.

We probably going to remove the `library/enterprise/minio` before merging as I've added it for testing to setup CI, that will be utilized later.

The multiple portals defined/commented out are there for some middleware testing, the one in the questions.yaml will be used and the others will be removed before merging.

Also there are some comments in the questions.yaml to discuss.


.github/ct-install-config/chart_schema.yaml
```diff
@@ -17,7 +16,8 @@ condition: str(required=False)
 keywords: list(str(), required=False)
 tags: str(required=False)
 maintainers: list(include('maintainer'))
-annotations: map(str(), str(), required=False)
+annotations:
+  title: str(required=True)
```
stavros-k, line 20: > As of [v3.3.2](https://github.com/helm/helm/releases/tag/v3.3.2), additional fields are not allowed. The recommended > approach is to add custom metadata in annotations.

https://helm.sh/docs/topics/charts/

enterprise/minio/1.0.0/questions.yaml
```diff
@@ -0,0 +1,318 @@
...
+                        show_if: [["type", "=", "hostPath"]]
+                    - variable: datasetName
+                      label: Dataset Name
+                      description: The name of the dataset to use for storage.
+                      schema:
+                        type: string
+                        show_if: [["type", "=", "ixVolume"]]
+                        required: true
+                        # immutable: true
+                        # Can we make this dynamic, so we can hide it?!
```
stavros-k, line 196: With the `immutable` in place, I can't edit it in the initial installation. 
Also is there any way to make the "dataset" name dynamic as it's in a list.

.github/ct-install-config/chart_schema.yaml
stavros-k: Key `title` also works, but using the recommended approach I think it might be safer in the long run?
In case Helm decide to hard fail if unexpected keys found?

enterprise/minio/1.0.0/questions.yaml
stavros-k: Latest nightly has solved the `immutable` issue.

library/ix-dev/enterprise/minio/values.yaml
```diff
@@ -0,0 +1,50 @@
+image:
+  repository: minio/minio
+  tag: RELEASE.2023-02-27T18-10-45Z
+  pullPolicy: IfNotPresent
+
+logsearchImage:
+  repository: minio/operator
+  tag: v4.5.8
+  pullPolicy: IfNotPresent
```
stavros-k, line 9: In this case that we have 2 images, how the `update_strategy` should look like?

stavros-k:
Depends on #1071

Qubad786 (commented):
Good work! @stavros-k you are missing `upgrade_info` file

library/ix-dev/enterprise/minio/questions.yaml
```diff
@@ -0,0 +1,345 @@
...
+        - variable: rootPass
+          label: Root Password
+          description: The secret key for the root user.
+          schema:
+            type: string
+            required: true
+            min_length: 8
+            private: true
+
+  - variable: minioRunAs
```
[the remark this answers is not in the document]
stavros-k: MinIO allows for it, But considering the amount of files that can be in those directories, we can't have any chown happening afterwards.

Probably make this immutable?

library/ix-dev/enterprise/minio/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+[MinIO](https://min.io) is a High Performance Object Storage released under Apache License v2.0.
+It is API compatible with Amazon S3 cloud storage service. Use MinIO to build high performance infrastructure
+for machine learning, analytics and application data workloads.
+
+> During the installation process, a container will be launched with **root** privileges. This is required
+> in order to apply the correct permissions to the MinIO data directory. Afterward, the `MinIO` container
+> will run as a **non**-root user (`568`).
+> Same applies to the `postgres` container. This will run afterwards as a **non**-root user (`999`).
+> On each upgrade, a container will be launched with **root** privileges in order to apply the correct
```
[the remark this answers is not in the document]
stavros-k: Not really, but because this is a backup dir we should be sure that we can write to it.
Probably apply the logic we did on qbit? Check parent dir before chowning?

library/ix-dev/enterprise/minio/questions.yaml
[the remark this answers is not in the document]
Qubad786: Makes sense

Qubad786 (approved):
Just one minor nit..please fix before merging. Good work ✅

Issue #1001, opened 2023-02-24
[Plex] Invalid value: 32400: provided port is already allocated


seanthewebber:
Version: TrueNAS-SCALE-22.12.1

## Steps to reproduce:

1.
- **Apps > Available Applications > Install**
- Install an instance of the `plex` application
- **Application name** is `plexmediaserver-1`
- [ ] Configure Host Network is false
- **Port to expose for Plex UI** is `32401`
2.
- **Apps > Available Applications > Install**
- Install an instance of the `plex` application
- **Application name** is `plexmediaserver-2`
- [ ] Configure Host Network is false
- **Port to expose for Plex UI** is `32402`
3.
- **Apps > Installed Applications >[plexmediaserver-1] > Start**
- Deploy is successful and status is `Active`
4.
- **Apps > Installed Applications >[plexmediaserver-2] > Start**
- Error: `Deploying 0/2`

`[EFAULT] Failed to update chart release: Error: UPGRADE FAILED: failed to create resource: Service "plexmediaserver-2-tcp" is invalid: spec.ports[0].nodePort: Invalid value: 32400: provided port is already allocated`

![Screenshot 2023-02-23 183152](https://user-images.githubusercontent.com/9794367/221063055-e4f4d7b1-a538-4949-9559-8e45cb611381.png)
![Screenshot 2023-02-23 183209](https://user-images.githubusercontent.com/9794367/221063079-819a8f75-3c70-4ae3-9156-db179a7af727.png)

```
Error: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 426, in run
    await self.future
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 461, in __run_body
    rv = await self.method(*([self] + args))
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 1186, in nf
    res = await f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/schema.py", line 1318, in nf
    return await func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/middlewared/plugins/chart_releases_linux/chart_release.py", line 553, in do_update
    await self.middleware.call('chart.release.helm_action', chart_release, chart_path, config, 'update')
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1386, in call
    return await self._call(
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1346, in _call
[...]

PR #1009, opened 2023-02-28
NAS-120551 / 23.10 / Plex: Set hostNetwork value in both true/false cases

stavros-k:
I had seen cases in the past where even if `hostNetwork: true` key was removed from a deployment, it will still have the effects as it was still set.

Explicitly setting it to true or false is safe, and could at least make sure that the hostNetwork should not be consumed.

It's also a potential fix for #1001, assuming `hostNetwork` was enabled at any point.


stavros-k:
Merging this and will circle back when more info are added to the initial issue.

PR #1011, opened 2023-02-28
Adapt charts CI and improve/fix common

stavros-k:


.github/workflows/common_library_tests.yaml
```diff
@@ -164,33 +167,14 @@ jobs:
...
-          ct install  --config .github/ct-install-config/ct-install.yaml \
-                      --charts library/common-test \
-                      --debug || (echo -e "\n\n--===PODLOGS===--\n\n" && \
-                                  cat /tmp/output.log && \
-                                  rm -f /tmp/output.log && exit 1)
-
-          kill $!
-          echo -e "\n\n--===PODLOGS===--\n\n"
-          cat  /tmp/output.log
-          rm -f /tmp/output.log
```
stavros-k, line 196: Initially was added because sometimes CI was faling to capture and display all logs.
hopping that the issue was coming from the specific action, added the above.
But this didn't make a difference. So removing it.

create_app.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    # Make sure directories exist
+    mkdir -p "$train/$app/$version"
+
+    # Copy files over
+    rsync --archive --delete  "library/$train/$app/" "$train/$app/$version"
+    # Rename values.yaml to ix_values.yaml
+    mv "$train/$app/$version/values.yaml" "$train/$app/$version/ix_values.yaml"
+    # Rename README.md to app-readme.md
+    cp "$train/$app/$version/README.md" "$train/$app/$version/app-readme.md"
```
Qubad786, line 39: This is not the case always?
App readme is supposed to be a smaller version of the app read me and readme being the detailed one

library/common-test/tests/container/envFixed_test .yaml
```diff
@@ -221,7 +221,7 @@ tests:
               - container-name1
       image: *image
       TZ: Europe/London
-      containerOptions:
+      resources:
```
Qubad786, line 224: The name of the file has redundant spaces?

library/common/1.0.0/docs/container/securityContext.md
```diff
@@ -57,3 +57,10 @@ workload:
               drop:
                 - ALL
 ```
+
+---
+
+Notes:
+
+When setting capabilities for containers, remember to **NOT** include `CAP_` prefix.
+For example, `CAP_NET_ADMIN` should be `NET_ADMIN`.
```
Qubad786, line 66: No reservations but might be more confusing for contributors?

library/common-test/tests/container/envFixed_test .yaml
stavros-k: nice catch!

library/common/1.0.0/docs/container/securityContext.md
stavros-k: That's how kubernetes wants them defined. Took me a while to remember it and documented it.

create_app.sh
stavros-k: Hmm, right! Lets have both then!

Qubad786 (approved):

library/common/templates/lib/chart/_names.tpl
```diff
@@ -27,7 +27,7 @@
 
   {{- $name := .name -}}
 
-  {{- if not (mustRegexMatch "^[a-z0-9]([a-z0-9-]){1,61}[a-z0-9]$" $name) -}}
+  {{- if not (mustRegexMatch "^[a-z0-9]([a-z0-9-]){0,61}[a-z0-9]$" $name) -}}
```
stavros-k, line 30:
```suggestion
  {{- if not (mustRegexMatch "^[a-z0-9]-?([a-z0-9]){0,61}[a-z0-9]$" $name) -}}
```
Don't allow `--`

library/common/templates/lib/chart/_names.tpl
```diff
@@ -27,7 +27,7 @@
 
   {{- $name := .name -}}
 
-  {{- if not (mustRegexMatch "^[a-z0-9]([a-z0-9-]){1,61}[a-z0-9]$" $name) -}}
+  {{- if not (mustRegexMatch "^[a-z0-9]-?([a-z0-9]){0,61}[a-z0-9]$" $name) -}}
```
stavros-k, line 30:
```suggestion
  {{- if not (mustRegexMatch "^[a-z0-9](-?[a-z0-9]){0,61}[a-z0-9]$" $name) -}}
```

library/common/templates/lib/chart/_names.tpl
```diff
@@ -27,7 +27,7 @@
 
   {{- $name := .name -}}
 
-  {{- if not (mustRegexMatch "^[a-z0-9]([a-z0-9-]){1,61}[a-z0-9]$" $name) -}}
+  {{- if not (mustRegexMatch "^[a-z0-9](-?[a-z0-9]){0,61}[a-z0-9]$" $name) -}}
```
stavros-k, line 30:
```suggestion
  {{- if not (mustRegexMatch "^[a-z0-9]([a-z0-9]-?|-?[a-z0-9]){0,61}[a-z0-9]$" $name) -}}
```

rehevkor5:
I'm confused about how this repo works:

- charts are referring to `file://../../../common/2207.0.0` as a dependency, but that doesn't exist. It would need to be `../../../library/common/2207.0.0/`
- That chart is inside another chart? There's a `library/common/Chart.yaml` at version 1.0.0, but directly inside it is the `2207.0.0` folder which is itself another chart. That seems unusual?

Do the dependencies need to be changed to `../../../library/common/2207.0.0/` so that `helm dependency update` works? Or is there some kind of build process that provides a chart at `../../../common/2207.0.0`?

Is 2207.0.0 intentionally packaged in a subfolder of the library/common chart? If so, how is that supposed to be used?

stavros-k:
Let me start by saying that the repo is currently under a stream of ongoing improvements. So it's possible to see things that does not make much sense for a while, but it's there usually to ensure backwards compat.

1. The path only needs to be resolvable while fetching the dependency, That's done while the app is in the `library/ix-dev/$train/$app`. The path should be resolvable from there.
2. The nested `2207.0.0` directory is indeed another common chart. But not used by the `1.0.0` common chart.
It will eventually get removed (After no apps using it). But for now it was placed there to keep backwards compat with existing apps and their CI.  Dependency paths `file://..` have been updated to point to the new location already 

Common chart generally looks into few subfolders when it's getting built/used. versioned dirs are not one of them.

PR #1012, opened 2023-02-28
NAS-120567 / None / Add qBittorrent Community App

stavros-k:
- [x] Rootless qBittorrent app, user can define any user id (except `0`)
- [x] A container (running as root) will spawn on each app startup to fix permissions on downloads/config directory.
  - [x] Contaienr runs with limited resources (1cpu/512ram)
  - [ ] Might want to limit this to config directory only? or only if the dir is empty? Let user decide?
- [x] Configurable web / bt ports (applies also when running in hostnetwork).
- [x] Allow for user defined env vars? Currently container allows to change any config option of qbit's config file with an env var. Tho there is zero validation for the input.

---

- [ ] Allow for user defined extra storage?


library/community/qbittorrent/templates/_qbittorrent.tpl
```diff
@@ -0,0 +1,96 @@
...
+    type: NodePort
+    targetSelector: qbittorrent
+    ports:
+      bt-tcp:
+        enabled: true
+        primary: true
+        port: {{ .Values.qbittorrent.network.btPort }}
+        nodePort: {{ .Values.qbittorrent.network.btPort }}
+        targetSelector: qbittorrent
+      bt-upd:
```
stavros-k, line 64:
```suggestion
      bt-udp:
```

Qubad786:
@sonicaj we should probably remove `deploy-charts` action as the chart is deployed/tested in a different pipeline already now?

library/ix-dev/community/qbittorrent/questions.yaml
```diff
@@ -0,0 +1,209 @@
...
+                        type: string
+                        required: true
+
+  - variable: qbitRunAs
+    label: ""
+    group: User and Group Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: user
```
Qubad786, line 59: This should not be allowed to change as well after the app has been deployed?

stavros-k: Technically can change. As long as the volumes are chownd again. 

I can see use case for it: 
A separate app that needs to access the same data, can't have its user changed. But you can then match qbit's user. 


Maybe we just add that info in the description/readme?

Qubad786: To confirm we would be changing ownership of the volumes again if run context changes?

stavros-k: Currently that's not the case. We only chown at startup. 
Meaning that if user/group is changed, user will have to do the chown manually

Qubad786: @stavros-k don't you think it should be either of the following:
1. Do not allow user to change run context
2. Change ownerships if run context changes

We could maybe just check if parent directory has different owner and then do chown on updates if we go with (2)?

stavros-k: If we go with (2), we might run into issues like parent dir having correct owner (in case user have manually changed it, or another app did) but sub-dir/files does not.
Resulting in a kinda broken state, which user in the end will have to chown manually. 

If we go with (1), can we later (if we need to) remove the `immutable` flag on upgrade?

Qubad786: @stavros-k we can remove that flag yes and i think (2) is not a huge problem either as the mismatch will only happen when the run context is changed and not otherwise so checking parent dir will be enough..

stavros-k: Added a new init container for addressing (2). Maybe we should change this as the default behavior on all init containers for permissions?

Qubad786 (approved):

PR #1023, opened 2023-03-06
NAS-120628 / 23.10 / Add IPFS app to community train

stavros-k:
* [x] Run as non-root (expect permissions container)
* [x] Permissions containers has its Resources is limited to 512M + 1vcpu
* [x] Configurable ports (applies when hostnet is enabled too)
* [x] Allow for user defined env vars
* [x] Apply port/headers configuration on each startup (Resources is limited to 512M + 1vcpu)
  * [ ] Maybe make headers optional?
* [ ] Allow custom config commands to be appended on the init script? Can be quite challenging for json payload. as users will be expected to correctly escape double quotes.

- - -
* [ ] Should this be named `kubo` or `ipfs`? Recently renamed to kubo (https://github.com/ipfs/kubo/issues/8959)
* [ ] Allow for user defined extra storage? Can't find something concrete if this app  needs extra.


catalog.json
```diff
@@ -646,5 +646,23 @@
...
+            "location": "/__w/charts/charts/community/ipfs",
+            "latest_version": "1.0.0",
+            "latest_app_version": "v0.18.1",
+            "latest_human_version": "v0.18.1_1.0.0",
+            "last_update": "2023-01-31 00:00:23",
+            "name": "ipfs",
+            "title": "Ipfs",
+            "icon_url": "https://ipfs.io/ipfs/QmVk7srrwahXLNmcDYvyUEJptyoxpndnRa57YJ11L4jV26/ipfs.go.png"
+        }
+    }
```
stavros-k, line 667: needs to go before merge, only here for testing

library/community/ipfs/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+  repository: ipfs/kubo
+  pullPolicy: IfNotPresent
+  tag: v0.18.1
+
+resources:
+  limits:
+    cpu: 4000m
+    memory: 8Gi
+
+ipfs_config:
```
Qubad786, line 11: Why use snake case when camel case is standard in helm?

library/community/ipfs/questions.yaml
```diff
@@ -0,0 +1,219 @@
...
+              - variable: hostPath
+                label: Host Path
+                schema:
+                  type: hostpath
+                  show_if: [["type", "=", "hostPath"]]
+                  immutable: true
+                  required: true
+
+  - variable: resources
+    label: ""
```
Qubad786, line 196: Missing label

library/community/ipfs/questions.yaml
```diff
@@ -0,0 +1,219 @@
...
+        - variable: hostNetwork
+          label: Host Network
+          description: |
+            Bind to the host network. It's recommended to keep this disabled.
+          schema:
+            type: boolean
+            default: false
+
+  - variable: storage
+    label: ""
```
Qubad786, line 119: Same here

library/community/ipfs/values.yaml
stavros-k: Will update in a bit

stavros-k: fixed

Qubad786 (commented):
We are missing upgrade strategies files so these apps won't get updates..
Also are we going to chown the storage paths?

stavros-k:
> We are missing upgrade strategies files so these apps won't get updates.. Also are we going to chown the storage paths?

I'll need help with the update strategies, at least until I get how it works.

From my testing `chown` is needed for both host path and ixVolumes, 
because the app does not runs as root.

Qubad786:
@stavros-k for apps which are already present in `charts` train like ipfs - we can copy the strategies directly and for apps which don't already have one,please create tickets for those..

Where are we doing the chown?

stavros-k:
> @stavros-k for apps which are already present in `charts` train like ipfs - we can copy the strategies directly and for apps which don't already have one,please create tickets for those..
> 
> Where are we doing the chown?

We chown `data` and `staging`

Qubad786:
And this happens only once on install?

stavros-k:
> And this happens only once on install?

Currently is at every startup, that was because I had to fix permissions before the init-config container runs. But permissions container was running after.

But with a little change I did in common before we merge it, I can now run it once at installation.

Qubad786:
I think it should be once on installation as chown can be expensive..

stavros-k:
> I think it should be once on installation as chown can be expensive..

Agreed.

library/ix-dev/community/ipfs/templates/_ipfs.tpl
```diff
@@ -0,0 +1,149 @@
...
+              command:
+                - ipfs
+                - dag
+                - stat
+                - /ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
+      initContainers:
+      {{- include "ix.v1.common.app.permissions" (dict "containerName" "01-permissions"
+                                                        "UID" .Values.ipfsRunAs.user
+                                                        "GID" .Values.ipfsRunAs.group
+                                                        "type" "install") | nindent 8 }}
```
Qubad786, line 53: Do you think having another workload which only targets the install time workload makes more sense?
It feels weird that we have a workload defined which will be actually running ipfs and then we have under init container some metadata which says that this is only applicable on installatino of chart?

stavros-k: So the `01-permissions` container is marked as `type: install`. Which common puts it between an if block of `.Release.IsInstall`.
Meaning it will only render at installation, upgrades will not render this at all.
ipfs container does not run as root, so it won't be able to do any chown.

The `02-init-config` is `type: init`, meaning it will get rendered (and run on every start of the app).
Which in this case is needed as it configures the ports.

Qubad786: I see..i was seeing this differently as being a pre-install job or something but obviously that is not the case

Qubad786 (approved):

library/ix-dev/community/ipfs/templates/_ipfs.tpl
stavros-k: Nop, it's just containers in the same Deployment.

Opted to not go for jobs in this case, as it's quick actions. And spinning a whole extra pod would require more configuration

PR #1031, opened 2023-03-14
Add chia to community train

stavros-k:


catalog.json
```diff
@@ -646,5 +646,24 @@
         }
     },
     "enterprise": {},
-    "community": {}
-}
\ No newline at end of file
+    "community": {
```
stavros-k, line 649: remove before merge

library/ix-dev/community/chia/Chart.yaml
```diff
@@ -0,0 +1,25 @@
+name: chia
+description: Chia is a modern cryptocurrency built from scratch, designed to be efficient, decentralized, and secure.
+annotations:
+  title: Chia
```
Qubad786, line 4: If `title` is not specified at top level dict,we capitalize the first letter of `name` but ideally we should be having it there jfyi

library/ix-dev/community/chia/values.yaml
```diff
@@ -0,0 +1,34 @@
...
+  farmer_address: ""
+  farmer_port: ""
+  ca: ""
+  full_node_peer: ""
+  additionalEnvs: []
+
+chiaStorage:
+  data:
+    type: ixVolume
+    hostPath: ""
```
Qubad786, line 28: Confused for this?

library/ix-dev/community/chia/Chart.yaml
stavros-k: Only reason moved it under annotation is because of this:
While technically Helm does not "hard" exit if there are other fields there, I'm not sure how it will be handled in future versions.

https://helm.sh/docs/topics/charts/#the-chartyaml-file
```
As of [v3.3.2](https://github.com/helm/helm/releases/tag/v3.3.2), additional fields are not allowed. The recommended approach is to add custom metadata in annotations.
```

library/ix-dev/community/chia/values.yaml
stavros-k: It's only there for reference on what questions.yaml have as default.
At install it will get overwritten from user entered values.

When common sees type: `ixVolume` will use the `datasetName` (and ignore `hostPath`). to retrieve the correct host path  from `ixVolumes` list.

If the type is `hostPath` it will ignore `datasetName` and use the supplied `hostPath`

stavros-k:
> LGTM white_check_mark Some minor nits..Also do you think we should be explicit about chown here for root?

We don't do any chown here, (only a chmod on the healthcheck script). 
App does runs as root. Which I assume it might do chown if the files do not have correct permissions. 

Tho the app does _not_ have the `CHOWN` linux capability, so it won't really chown. But it should be able to modify files which will change the owner.

Qubad786 (approved):

PR #1033, opened 2023-03-16
NAS-121062 / 23.10 / Add github workflow for validating apps under developement dir

sonicaj:


sonicaj:
@Qubad786 it is ready now

Qubad786 (approved):

PR #1034, opened 2023-03-16
NAS-121069 / None / Add homebridge to `community` train

stavros-k:
- App runs as `root`
- auto device discovery uses mDNS. Which means
  - hostnetwork is needed
  - mDNS must be disabled on the host.
  - other solution is hostnetwork disabled and enabled mDNS on the host, but external interface used. Will need manual configuration from the user on the config file (available via the in-app gui) to bind the external interface for mdns (https://github.com/homebridge/homebridge/issues/1957#issuecomment-410505653)


Qubad786:
@stavros-k is this up for review?

library/ix-dev/community/homebridge/README.md
```diff
@@ -0,0 +1,18 @@
+# Homebridge
+
+- This container runs as `root` user.
+
+If host networking is enabled, you must disable mDNS on the host.
```
Qubad786, line 5: @anodos325 would this be an issue?

yocalebo: Time Machine functionality is lost if mDNS is turned off on host as well as the convenience MacOS users have by truenas magically showing up in their network bar

Qubad786: Is this acceptable?

Qubad786: @stavros-k let's not allow auto device discovery then and re-consider when we get requests etc?

stavros-k: Updated, note that it still relies on #1124 for the external Interfaces

anodos325:
Do you happen to know how it needs to use mDNS? Is it generating a special SRV record?

anodos325:
https://github.com/homebridge/ciao#notice-on-native-mdns-responders

Looks like they have their own wiki entry which more or less repeats my warning. It does however look like you could rewrite our mDNS implementation to use dbus because it looks like homekit may be able to communicate with avahi via it (though not sure how this works in apps context.

anodos325:
One other alternative to getting mDNS config from homebridge to system avahi is to write an srv file into the avahi dir. Do you happen to know what the SRV records are they're trying to write?

anodos325:
We already have a generic framework to generate mDNS SRV records
https://github.com/truenas/middleware/blob/5e730d996d1a2350775b51e8a3a3434dbf5782f2/src/middlewared/middlewared/etc_files/local/avahi/avahi_services.py#L27

Why not just expand it to include homebridge info if someone is using this app?

anodos325:
> If it's gonna be expanded it might worth making it a bit more generic. As other apps might need to use mDNS in the future.

I'm okay with creating a db table for mDNS services and having apps add what they need to it.

stavros-k:
@Qubad786 Can we have an interface for that in a form of a $ref or something?

PR #1039, opened 2023-03-20
Remove test train

sonicaj:


sonicaj:
@stavros-k already made the changes for test train :P

PR #1040, opened 2023-03-20
NAS-121113 / 23.10 / Retire charts deploy action

Qubad786:
## Context

Charts are validated now already with `ct-install` and there is no need to deploy a separate cluster to install/run apps in github actions.


sonicaj (approved):
Please remove the script from catalog_validation too please. Thanks!

PR #1049, opened 2023-03-22
Move Pihole to ix-dev and fix DHCP/DNS

stavros-k:


stavros-k:
Last commit is fixes for DNS/DHCP + hostNetwork (originally worked on #968).

- WebUI port is now always set to port specified in from the user (questions.yaml). Which is > 1024. 
  -  That makes sure that turning hostnetwork on the app won't break SCALE webui.
- When hostnetwork is enabled, hides from UI port configuration for DHCP and DNS, as it does not apply.
  - (Container will bind 53/67 on the host in this case)
  - Also `NET_ADMIN` cap is added on hostNet, so it can actually bind those ports.
- Added DHCP required vars (start, end, gateway) when dhcp is enabled)

Qubad786 (approved):

PR #1060, opened 2023-03-27
NAS-121233 / 23.10 / fix(pihole): hostnetwork toggle not showing and dhcp

stavros-k:
Removed services and hostnetwork option and hardcoded it to true. Reason is that since NodePort can't bind port 53/67 on the host, devices outside the k8s cluster will never reach pihole. So it needs to bind host's ports for that to work.

The only case that you might not need to bind host's 53 port, is if you use pihole as a DNS server behind another DNS server.
Which it might have the option to set a different port.


stavros-k:
We can also leave the "hostnetwork" option in the GUI, but make it "required" so user explicitly enables it.

stavros-k:
Yep,Expected. Needs a rebase!

yocalebo (approved):
SHIP IT

PR #1061, opened 2023-03-27
Add adguard-home to community train

stavros-k:


library/ix-dev/community/adguard-home/questions.yaml
```diff
@@ -0,0 +1,181 @@
...
+    protocols:
+      - "$kubernetes-resource_configmap_portal_protocol"
+    host:
+      - "$kubernetes-resource_configmap_portal_host"
+    ports:
+      - "$kubernetes-resource_configmap_portal_port"
+    path: "$kubernetes-resource_configmap_portal_path"
+
+questions:
+  # - variable: adguardRunAs
```
Qubad786, line 24: Do we need this?

stavros-k:
```
# FIXME: It might be able to run rootless, probably blocked by:
# https://github.com/AdguardTeam/AdGuardHome/issues/4681
```
Don't know yet if when the issue above is solved it will be able to run as arbitrary user.

PR #1065, opened 2023-03-29
NAS-121235 / 23.10 / Add gitea to community train

stavros-k:
- Runs rootless
- Only the required configuration is plumbed, for postgres and optional for certs. Rest of the config, can be either added as additional env vars or via editing the app.ini in the config volume.


stavros-k:
Depends on #1071

library/ix-dev/community/gitea/questions.yaml
```diff
@@ -0,0 +1,315 @@
...
+        - variable: user
+          label: User ID
+          description: The user id that Gitea will run as.
+          schema:
+            type: int
+            min: 1
+            # FIXME: See values.yaml
+            hidden: true
+            editable: false
+            default: 1000
```
Qubad786, line 70: Why not 568?

Qubad786 (approved):

PR #1072, opened 2023-03-31
NAS-121272 / 23.10 / Adds prometheus to `charts` train

stavros-k:
- Runs rootless.
- While web can be configured with basic auth and/or tls, those features not added currently for 2 reasons
  - TLS is still experimental, can only be configured via a config file. which also use will need to be able to modify.
    and using `sed` and/or `grep` is not ideal. Once TLS is beta/stable we can at least mount the certs but will be up to the user
    to configure it. Also portal + probes will need a checkbox to switch to https once that happens.
  - Basic auth currently is also covering the healthcheck endpoints. there is some WIP for that upstream, to exclude that 
    endpoints (basic auth is also added in the config file, so we can't know the credentials to pass them to the probes)
- Creates an empty configuration file as it's required by prometheus in order to start. But beyond that is up to the user to 
  configure targets for scraping.
- Some commonly used arguments are added to the GUI, but also given the ability to pass custom additional flags and/or env's


Qubad786 (approved):

Qubad786 (changes requested):
This should point to community ?

stavros-k:
> This should point to community ?

Nop it's under Official train in the list

Qubad786 (approved):

PR #1077, opened 2023-04-04
Make init container for permissions more flexible and apply to qbit

stavros-k:
- Adapts "permission" template to have 2 modes, `always` and `check`.
`check`: checks if the parent dir have the correct permissions, if not, `chown's`
`always`: does not check, chown's always

- Applies this template to qBittorrent.
- Removes qBittorent from `charts` train (second commit)


Qubad786 (approved):
Will fix the CI, looks like `last_update` is the problematic field probably..

PR #1078, opened 2023-04-04
NAS-121334 / 23.10 / Add ElasticSearch chart to `charts`

stavros-k:
- Runs rootless (uid, gid is not configurable in this case)
- http or https depending if user selects a cert or not


library/ix-dev/charts/elastic-search/questions.yaml
```diff
@@ -0,0 +1,176 @@
...
+    group: Elastic Search Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: password
+          label: Password
+          description: The password for the Elastic Search (elastic) user.
+          schema:
+            type: string
+            default: ""
```
Qubad786, line 36: JFYI defaults do not need to be specified for attrs which are `required` and user should be specifying values for them always

library/ix-dev/charts/elastic-search/questions.yaml
```diff
@@ -0,0 +1,176 @@
...
+            required: true
+        - variable: nodeName
+          label: Node Name
+          description: The name of the Elastic Search node.
+          schema:
+            type: string
+            default: elastic_search_node
```
Qubad786, line 43: Probably better to quote and below as well just for consistency for string schema?
Might bite us in future with something being considered as boolean/integer and not string or something..

Qubad786 (approved):

PR #1082, opened 2023-04-06
NAS-121366 / 23.10 / Adds sonarr to `charts` train

stavros-k:
- Runs rootless
- Can add additional storage (Usually /tv or /downloads)


Qubad786 (approved):

stavros-k:
Partial covers #1094

PR #1084, opened 2023-04-06
NAS-121368  (Minio)Add a default storage item and description on storage type 

stavros-k:
While a validation is performed on the chart side to prevent installing without storage, this PR:
Adds a default storage item, since at least 1 is required.
Also shortly explains what an ixVolume and a host path type is

### DIFF
--- a/library/ix-dev/enterprise/minio/questions.yaml
+++ b/library/ix-dev/enterprise/minio/questions.yaml
@@ -187,9 +187,9 @@ questions:
                   default: ixVolume
                   enum:
                     - value: hostPath
-                      description: Host Path
+                      description: Host Path (Path that already exists on the system)
                     - value: ixVolume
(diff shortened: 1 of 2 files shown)


library/ix-dev/enterprise/minio/questions.yaml
```diff
@@ -175,6 +177,9 @@ questions:
             attrs:
               - variable: type
                 label: Type
+                description: |
```
sonicaj, line 180: I think in the description of enum we should specify this as well

sonicaj (approved):
If we have the same format on other apps, let's please change this there as well in a different PR. Thanks!

sonicaj:
@stavros-k also moving on please use the ticket number for your branch name - bug clerk will take care of the integration with jira then.

PR #1089, opened 2023-04-07
NAS-121379 / 23.10 / Adds radarr to charts train

stavros-k:


library/ix-dev/charts/radarr/questions.yaml
```diff
@@ -0,0 +1,225 @@
...
+    schema:
+      type: dict
+      attrs:
+        - variable: user
+          label: User ID
+          description: The user id that Radarr will run as.
+          schema:
+            type: int
+            min: 1
+            default: 568
```
Qubad786, line 72: Should it be immutable?

stavros-k: It's the same thing as qbit.
The permissions container checks the parent dir here before chown

Qubad786 (approved):

stavros-k:
Partial covers #1094

PR #1106, opened 2023-04-10
NAS-121410 / 23.10 / Avoid having the same key defined twice

stavros-k:
When defining custom env's the `env` key would be defined twice.
This makes sure key only appears once, also fixes the `sources` links


stavros-k:
CI Failure is "expected" here, as it runs also an upgrade from the previous version.

Previous version includes a wrong variable name `RADARR__AUTHENTICATION_METHOD` which was set to `some-long-api-key` which is not valid value.

No user is affected as this is only values used in CI (ci/extra-values.yaml).

Qubad786 (approved):

PR #1117, opened 2023-04-12
NAS-121454 / 23.10 / Add nginx-proxy-manager to `community` train

stavros-k:
- Requires root and specific PUID/PGID to run.

⚠ If hostnetwork is enabled it will bind 80,81,443. 
Maybe we want to disallow host network?
I can't think of any reason to use hostnetwork here. A portforward to the NodePort Service should be enough


library/ix-dev/community/nginx-proxy-manager/templates/_npm.tpl
```diff
@@ -0,0 +1,103 @@
...
+                - SETUID
+                # Needed for: s6-applyuidgid: fatal: unable to set supplementary group list: Operation not permitted
+                - SETGID
+                # Needed for: Used in some s6-overlay startup scripts
+                - CHOWN
+                - DAC_OVERRIDE
+                # Needed for: Nginx Service
+                - FOWNER
+          fixedEnv:
+            PUID: 1000
```
stavros-k, line 33: Currently it only works for a few PUIDs, there are some upstream bugs lately. Once it's stabilized a bit, we can exposed it to the users.
(For example 568 does not work, but it mostly works for PUIDs 1000+)

Qubad786 (commented):
Yes let's not allow hostnetwork please..would be very undesirable to lose SCALE UI access because we allowed an app which forced used system ports required by system

t-abraham:
I am sorry to bring this stupid question. I am still learning TrueNAS. As mentioned by @stavros-k "I can't think of any reason to use hostnetwork here. A portforward to the NodePort Service should be enough". can anyone please tell me how to do that? I wish to use nginx-proxy-manager to proxy local domains to different services which are running in the apps of TrueNAS...

to my understanding, It is required for nginx-proxy-manager to be listening to port 80, 443, for the http, https services and other ports which are below 9000 (if i have streams setup in the nginx-proxy-manager)

PR #1118, opened 2023-04-12
NAS-121456 / 23.10 / Adds field for emptyDir `sizeLimit` to ix-chart

stavros-k:
Adds a field for sizeLimit on memory backed volumes. It's an optional field

### DIFF
--- a/library/ix-dev/charts/ix-chart/questions.yaml
+++ b/library/ix-dev/charts/ix-chart/questions.yaml
@@ -494,6 +494,7 @@ questions:
                   Format: 100Mi, 1Gi, 2Gi etc
                 schema:
                   type: string
+                  valid_chars: "^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$"
 
   # Volumes
   - variable: volumes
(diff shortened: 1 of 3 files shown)


library/ix-dev/charts/ix-chart/questions.yaml
```diff
@@ -487,6 +487,14 @@ questions:
                 schema:
                   type: path
                   required: true
+              - variable: sizeLimit
```
Qubad786, line 490: Let's add a reasonable default?

library/ix-dev/charts/ix-chart/questions.yaml
```diff
@@ -487,6 +487,14 @@ questions:
...
+                label: "Size Limit"
+                description: |
+                  Optional - Size of the memory backed volume.</br>
+                  Format: 100Mi, 1Gi, 2Gi etc
+                schema:
+                  type: string
+                  valid_chars: "^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$"
```
Qubad786, line 497: I wonder if we can somehow have unit tests for such regex matches?

library/ix-dev/charts/ix-chart/questions.yaml
stavros-k: I think 2Gi is okay for most systems?

Qubad786: Maybe 1Gi?

Qubad786: Probably even less then that might be better..

Qubad786 (approved):
Please fix the conflict otherwise LGTM ✅

PR #1122, opened 2023-04-13
NAS-121476 / 23.10 / Add tailscale to `community` train

stavros-k:
In my testing with userspace enabled, I could access Scale WebUI/SMB and with correct `Advertise Routes` I could access devices in Scale's LAN.
(ping was not working, due to running with an unprivileged user and busybox required root)


library/ix-dev/community/tailscale/questions.yaml
```diff
@@ -0,0 +1,158 @@
...
+          label: Hostname
+          description: |
+            The hostname for Tailscale Node.</br>
+            Only lowercase letters, numbers, and hyphens are allowed.</br>
+            Same as `--hostname` flag.
+          schema:
+            type: string
+            default: "truenas-scale"
+            required: true
+        - variable: advertiseRoutes
```
Qubad786, line 36: Just to be sure these are not always required, right?

Qubad786 (approved):

library/ix-dev/community/tailscale/questions.yaml
stavros-k: Correct, routes are only required when running in userspace and/or when you want to route devices in the same lan as the host.
It's needed 95% of the cases, but not always.

PR #1124, opened 2023-04-13
NAS-121481 / 23.10 / Set nvidia caps to void when no gpu is passed, also adds `render` group when a gpu is selected and other small fixes

stavros-k:
There are some apps that pre-set `NVIDIA_VISIBLE_DEVICES` to either `all` or other values.
This can lead to issues on systems that have legacy GPUs that are not supported.

This probably won't be an issue on Cobia as it shouldn't use the nvidia runtime without a supported GPU, but we can already set the `NVIDIA_VISIBLE_DEVICES` to `void`, so even with nvidia runtime being used, it wouldn't cause any issues.

This shows a case.
https://github.com/truecharts/charts/issues/7955

Additionally adds `render` as `supplementalGroup` when a GPU is selected

Also;
- added some tests to init container that was missing)
- fixed a small edge case on when type casting and then comparing
- fixed a missing nest level on external interfaces
- added some missing docs for dnsConfig


Qubad786 (approved):
I suppose this does not affect ix-chart?

stavros-k:
> I suppose this does not affect ix-chart?

Nope, only the new common chart

PR #1149, opened 2023-04-20
NAS-121592 / 23.10 / Add mumble to `community` train

stavros-k:
- App runs as a **non**-root user. (1000) but is not configurable.
- Allows for easy certificate setup
- Provides few common config options on the GUI.
- Allows for additional env variables to fully customize the server


Qubad786:
Is this up for review?

stavros-k:
> Member

yes

library/ix-dev/community/mumble/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+annotations:
+  title: Mumble
+type: application
+version: 1.0.0
+apiVersion: v2
+appVersion: 'v1.4.230'
+kubeVersion: '>=1.16.0-0'
+maintainers:
+  - name: truenas
+    url: https://www.truenas.com/
```
stavros-k, line 12:
```suggestion
    url: https://www.truenas.com/
    email: dev@ixsystems.com
```

yocalebo:
Check with @anodos325 before merging this. UID starting at 1000 could collide with existing local users.

library/ix-dev/community/mumble/questions.yaml
```diff
@@ -0,0 +1,189 @@
...
+    group: Mumble Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: serverPassword
+          label: Server Password
+          description: The password for the Mumble Server.
+          schema:
+            type: string
+            default: ''
```
Qubad786, line 23: There is no need to provide an empty string here and `empty` should be false as this has been provided and will pass validation

Qubad786: Same as below

library/ix-dev/community/mumble/templates/_mumble.tpl
```diff
@@ -0,0 +1,119 @@
...
+            startup:
+              enabled: {{ .Values.ci }}
+              type: tcp
+              port: "{{ .Values.mumbleNetwork.serverPort }}"
+      initContainers:
+      {{- include "ix.v1.common.app.permissions" (dict "containerName" "01-permissions"
+                                                        "UID" 1000
+                                                        "GID" 1000
+                                                        "mode" "check"
+                                                        "type" "init") | nindent 8 }}
```
Qubad786, line 49: I think this meant that each time container start it will check parent and then chown..is that required?

stavros-k: It will only chown if the uid in the parent have changed.
If we assume user won't change the permissions on the hostpath via UI/shell or other means, it's not needed.
Let's change it to only run on install once, considering that they UID is hardcoded and not configurable?

Qubad786: Sounds good to me ✅

stavros-k: Done

Qubad786 (approved):

PR #1158, opened 2023-04-25
NAS-121666 / 23.10 / Add minecraft to `community` train

stavros-k:
App runs as root.

Considering the amount of combinations of images, types and versions that exist out there, it's not possible to test them all on CI.
A basic "suite" of tests is added, using every image with VANILLA type and LATEST version (expect on some images that only support up to a specific version, those set to the latest version supported).

Commonly used options are exposed in the GUI and should not be blocking for any combination.
There is still an option to add additional env variables if needed (eg in cases of a non VANILLA type, it might need some api keys)


library/ix-dev/community/minecraft/upgrade_strategy
stavros-k, line 1: How would this look like for multiple images?

Qubad786: Please go through https://github.com/truenas/catalog_update/blob/master/README.md

Qubad786 (approved):
LGTM apart from `upgrade_strategy`..we can also merge this and remove the upgrade strategy as long as you create another ticket and comment the ticket number here

stavros-k:
Opened a ticket to adjust update_strategy for more than 1 images
https://ixsystems.atlassian.net/browse/TNCHARTS-162

Issue #1125, opened 2023-04-13
[Syncthing] Running multiple instances of the syncthing app not possible


PStibbons2K:
Dear all, 

there seems to be an issue with the SyncThing chart - the configurable ports from the TrueNAS GUI are apparently not properly used during the pod deployment or something else is wrong. I ran into this issue trying to run multiple instances of the syncthing application, resulting in errors during startup due to ports already in use.

Steps to reproduce:
 
- Create a named app instance (e.g. syncthing-user1) using the default exposed ports (20910, 20978)
- Create a second app instance (e.g. syncthing-user2) using different exposed ports (e.g. 21910, 21978)
- try starting both apps

The first app is starting fine, the exposed ports are usable and there is a dedicated pod for the app. The second instance fails with this error, even if there is a dedicated pod for the second instance:

```
2023-04-13 19:20:10 0/1 nodes are available: 1 node(s) didn't have free ports for the requested pod ports. preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod.
2023-04-13 19:14:56 Scaled up replica set syncthing-user2-b754c9885 to 1
2023-04-13 19:14:56 Created pod: syncthing-user2-b754c9885-tl6j9
```

As far as i understood the kubernetes concept. each pod should get its own IP adress and ports should be only blocked by containers within the same pod?

The status window of the running app shows additional ports - the syncthing default ports (8384, 22000).

What am i missing? Is there a way to run multiple instances of the same app?

Thanks and kind regards,

Martin

TrueNAS Version: TrueNAS-SCALE-22.12.2

PStibbons2K:
A bit more log output, seems the standard logging levels are not sufficient to get more information on the issue:


**k3s_daemon.log**
```
Apr 14 08:11:44 truenas k3s[5143]: I0414 08:11:44.623588    5143 event.go:294] "Event occurred" object="ix-syncthing-user2/syncthing-user2" fieldPath="" kind="Deployment" apiVersion="apps/v1" type="Normal" reason="ScalingReplicaSet" message="Scaled down replica set syncthing-user2-66444c94d9 to 0 from 1"
[...]

PR #1166, opened 2023-04-29
NAS-121749 / 23.10 / Apply a partial fix for hostnetwork toggling

stavros-k:
This is a partial fix for an issue that appears after enabling hostNetwork and then disabling it.
I did a small "writeup" here, in a k3s discussions. 
https://github.com/k3s-io/k3s/discussions/6289#discussioncomment-5760503 (Check my comment)

(This is not a 100% fix for apps that have TCP and UDP ports with matching number. Check the link above)
Partially solves #1125
Should also help with #1001

> This can't be really fully unit tested fully as the generated manifests from Helm does not contain any hostPort values when hostNetwork is enabled. Those are added by k8s

Also note that apps that uses the same containerPort for multiple protocols (like Syncthing + port 22000). Cannot be "fixed".

### DIFF
--- a/library/common-test/Chart.yaml
+++ b/library/common-test/Chart.yaml
@@ -16,3 +16,4 @@ dependencies:
 maintainers:
   - name: truenas
     url: https://www.truenas.com/
+    email: dev@ixsystems.com
--- a/library/common-test/tests/container/ports_test.yaml
+++ b/library/common-test/tests/container/ports_test.yaml
@@ -61,6 +61,7 @@ tests:
             ports:
               - name: port-name
                 containerPort: 80
+                hostPort: null
                 protocol: TCP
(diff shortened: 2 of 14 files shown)


stavros-k:
https://github.com/k3s-io/k3s/issues/7381
Created an issue in k3s

stavros-k:
https://github.com/kubernetes/kubernetes/issues/117689
https://github.com/kubernetes/kubernetes/issues/105610

stavros-k:
Applied the "fix" to few more apps that people usually play with hostNetwork too.

yocalebo:
@sonicaj @Qubad786 ping

Qubad786:
@stavros-k is this up for review?

Qubad786 (approved):
Please update your branch from master before

Issue #1167, opened 2023-04-29
Community chart suggestion : Overseerr


ianmarmour:
Service Name: Overseerr
Website: https://overseerr.dev/
Description:
> Overseerr is a request management and media discovery tool built to work with your existing Plex ecosystem.

GitHub: https://github.com/sct/overseerr
DockerHub: https://hub.docker.com/r/sctx/overseerr

ianmarmour:
I'm going to give my best attempt at following the `contributing.md` here (I say my best attempt because parts of the contribution guide don't really make sense to me as an outsider of ixSystems). Anyway here we go,

Which app is being proposed to be added to the catalog ? 
> Overseer

How do you think it is useful for users ?
> For users that are running Plex + Radarr/Sonarr on their TrueNAS Scale system it gives a safe way to expose a web UI to the internet for requests.

Who will be maintaining the new application ?
> The external application/container is maintained by the `Overseerr` team while I'm happy to maintain this community contribution for TrueNAS for now I think longer term reaching out to them to see if they would be interested is in order.

PR #1169, opened 2023-04-30
Add overseeer to the `community` train

ianmarmour:
> This application submission request still needs to be approved by `ixSystems`.

This PR adds a new application `overseeer` to the community train, I was able to deploy this locally on `TrueNAS-SCALE-22.02.0`. 

I mostly based this work off of existing charts that were already created in the `community` library so there may be extraneous configuration that is unnecessary (though I think I got most of it). I also had a bit of a hard time following the requirements in the `contributing.md` specifically around helm testing I'll record those here for feedback and potential improvements to the contribution guide,

> 1. Create a github issue [here](https://github.com/truenas/charts) which should outline the following points:

Would be awesome if this was a GitHub issue template where we could ask for all the required fields.

> 3. After (2) approval, create a PR [here](https://github.com/truenas/charts/pulls) adding the application to community train making sure that it complies with the [Official Catalog Application Standards](https://github.com/truenas/charts/blob/master/contributing.md#application-standards)

How long should we wait for approval? How should we request approval? I think there's some unknowns in this process still that could use work. Beyond that the links to existing helm tests are broken currently the only helm tests I see are generic common helm tests? Also some basic guidance on helm testing might be nice, what libraries to install etc... would be a nice improvement.

> 4. The application would be added to the community train after passing (3).

After passing what? Is there some test suite that we should be executing for validation locally? If it's a part of the CI system getting it exposed remotely would be pretty sweet just so we can run those tests. I know personally to do E2E testing I needed to run the `create_app.sh` script and then add a new catalog to a spare trunas box I had laying around.


ianmarmour:
This PR is related to,

#1167

community/overseerr/1.32.5/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
```
ianmarmour, line 1: I'm assuming everything in the community/* folder should not be included in the PR as it's generated from the `create_app.sh` script and is probably built during a CI process somewhere?

PR #1171, opened 2023-05-01
NAS-121758 / 23.10 / Change default NodePort for storj

stavros-k:
With the "mismatch" of external and internal ports users was not sure what ports they have to forward in their network in order to get Storj app to work at its fullest capacity. Storj web UI was mentioning that user have to port foward port 28967, but in reality user should forward the port configured in the TrueNAS UI (NodePort).

Matching nodePort with containerPort will at least make easier for new users to forward the correct port


yocalebo:
@sonicaj @Qubad786 ping

Qubad786 (approved):

PR #1189, opened 2023-05-07
Remove 20 char limit for the vaultwarden admin token

jlsjonas:
Hashed tokens (as per recommendation) will exceed the imposed 20 char limit.

Also see [Enabling admin page - Secure the `ADMIN_TOKEN`](https://github.com/dani-garcia/vaultwarden/wiki/Enabling-admin-page#secure-the-admin_token)

### DIFF
--- a/library/ix-dev/community/vaultwarden/Chart.yaml
+++ b/library/ix-dev/community/vaultwarden/Chart.yaml
@@ -4,7 +4,7 @@ description: Alternative implementation of the Bitwarden server API written in R
 annotations:
   title: Vaultwarden
 type: application
-version: 1.0.6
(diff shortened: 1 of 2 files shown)


stavros-k (changes requested):
I think this was intended to be `min`. But now it's not easy to enforce that.
Agreed on removing
Please, make changes **only** in the vauiltwarden under library/ix-dev/community/vaultwarden and  also bump `version` in Chart.yaml

stavros-k (approved):

Qubad786 (approved):

Qubad786 (approved):

PR #1210, opened 2023-05-18
NAS-122030 / 23.10 / Adds check to only render hostPath key if it has a value

stavros-k:
When chart was rendering it defaulted hostPath to empty value `""`.
This results to an error on upgrade due to some relative path checks being done in the middleware.
Resolving `""` to `.` and raising and error when hostPath Validation checks are enabled.

To avoid that, added a check that a non empty value is defined for hostPath, and render it only then.


Qubad786 (commented):
You will need a migration to fix existing user's values..let's please add that

stavros-k:
> You will need a migration to fix existing user's values..let's please add that

I'll look into the migration, 
I'll also revert the changes in the `tpl` files as this keys are not on the schema(questions.yaml) and are only visible by helm (which already performs those checks in the common). The issue comes only from the values.yaml (ix_values.yaml).

Tested on a live system, that only from ix_values.yaml needs to be removed.

Qubad786:
@stavros-k lgtm..however can you resolve conflicts please and i can re-review?

stavros-k:
> @stavros-k lgtm..however can you resolve conflicts please and i can re-review?

Done!

Qubad786 (approved):

PR #1216, opened 2023-05-23
NAS-122095 / 23.10 / Increase startup probe timeout

### LINKED ISSUE #1212: Nginx Proxy Manager do not start on bare-metal HDD based pools, but works on virtual disks and SSDs
I got a strange one for you today.

Nginx Proxy Manager seem to be unable to start, for an unknown reason. But specifically on HDD based pools, not virtual one (from a virtualized TrueNAS), nor SSDs based one.

I've setup 3 different pools as follow:

- 2x4To Ironwolf, mirrored, unencrypted
- 2x4To Ironwolf, mirrored, encrypted
- 2x500Go WD Red SSDs, encrypted

For each pool I did the following, from the freshly created pool.

- Set K3S to said pool, from the UI
- Start a new nginx-proxy-manager deployment
- Wait for it do come online and accessible

On the SSDs array, no problem. On both HDDs arrays, nginx-proxy-manager refuse to start and go on a boot loop.

No other deployment, be it from the official or community trains, seem to be affected.

Should I call an exorcist for my HDDs? ^^"

stavros-k:
This application runs chown on a number of directories at each startup, which can take quite a while on slower drivers (eg HDDs).
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/2a06384a4aa597777931d38cef49cf89540392e6/docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/30-ownership.sh#L7

Increasing the startup timeout from `5 minutes` to `10 minutes` should be safe.
(probe hits every 5 seconds)

FIxes #1212


Qubad786 (approved):

ejalal:
Hey all, is there a reason why it does chown 1000:1000 ?
The user/group 1000 doesn't exist in v23.10, and it creates problems
I'd rather have change the owner to 568:568 which is the apps user/group (or root if it needs to run as root)

PR #1227, opened 2023-05-30
NAS-122154 / 23.10 / Add rsyncd to `community` train

stavros-k:
- It should have feature parity with the host rsync service.
- It runs as root

TODO: Update container to stable after https://github.com/truenas/containers/pull/3 is merged


library/ix-dev/community/rsyncd/ci/hostnet-multiple-values copy.yaml
```diff
@@ -0,0 +1,59 @@
+rsyncNetwork:
```
Qubad786, line 1: You have a space in the file name..is that okay?
I would assume yes - it's just that it's not common to see that :D

library/ix-dev/community/rsyncd/questions.yaml
```diff
@@ -0,0 +1,235 @@
...
+    description: Configure Resources for Rsync
+
+questions:
+  - variable: rsyncConfig
+    label: ""
+    group: Rsync Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: logToStdout
```
Qubad786, line 18: Do you see any value to logging this to a file?
And if it is to be logged to some file then we should be doing something custom which exposes a convenient method for a user to know that his logs are being saved at X location

library/ix-dev/community/rsyncd/templates/_validation.tpl
```diff
@@ -0,0 +1,83 @@
...
+  {{- end -}}
+{{- end -}}
+
+{{- define "rsync.module.validation" -}}
+  {{- $mod := .mod -}}
+  {{- if not $mod.name -}}
+    {{- fail "Rsync - [Module Name] is required." -}}
+  {{- end -}}
+
+  {{- if not (mustRegexMatch "^[a-zA-Z0-9]+([_-]*[a-zA-Z0-9]+)+$" $mod.name) -}}
```
Qubad786, line 30: We should have this in questions schema too please

library/ix-dev/community/rsyncd/values.yaml
```diff
@@ -0,0 +1,19 @@
+# FIXME: Switch to stable tag
+image:
+  repository: ixsystems/rsyncd
+  pullPolicy: IfNotPresent
+  tag: 'unstable'
```
Qubad786, line 5: I assume you want to change this to `1.0.0`

library/ix-dev/community/rsyncd/templates/_validation.tpl
stavros-k: Added

library/ix-dev/community/rsyncd/values.yaml
stavros-k: Correct

library/ix-dev/community/rsyncd/ci/hostnet-multiple-values copy.yaml
stavros-k: There isn't an issue, but good catch! fixed!

library/ix-dev/community/rsyncd/questions.yaml
stavros-k: Updated the wording, this is mostly need to be togglable in case someone want to use syslog.
(If log file is set, syslog is ignored according to rsyncd docs).

Qubad786: I don't follow?
In this case the logs will not be redirected to some file instead?

stavros-k: You are correct. Mixed it up with **r**syslog.

In this case, we can either 
1. Remove the toggle for now until we have an actual use case for logging, 
2. Allow the user to add additional volumes (and combine those with the aux parameters). As there are multiple ways to configure logging and some options can't exist at the same time. (eg having log file set, syslog is ignored.)

Qubad786: Let's remove it please..we should be making the process easier - so if we were to add it, there should be automatic configuration of volume here etc and i think as we don't have an actual usage atm we should just leave it.

stavros-k: Done

Qubad786 (approved):
Please update the image tag before merging..thanks!

library/ix-dev/community/rsyncd/values.yaml
```diff
@@ -0,0 +1,18 @@
...
```
stavros-k, line 5:
```suggestion
  tag: '1.0.0'
```

PR #1230, opened 2023-05-31
NAS-122254 / 23.10 / Update adguard home to v0.107.29

xxyzz:
[v0.107.27](https://github.com/AdguardTeam/AdGuardHome/releases/tag/v0.107.27) fixes security vulnerabilities and should be updated. I guess your auto update bot is not working for this docker image because it doesn't have the `upgrade_info` JSON file.


library/ix-dev/community/adguard-home/upgrade_strategy
```diff
@@ -6,14 +6,14 @@ import sys
 from catalog_update.upgrade_strategy import semantic_versioning
 
 
-RE_STABLE_VERSION = re.compile(r'v[0-9]+\.[0-9]+\.[0-9]+')
+RE_STABLE_VERSION = re.compile(r'v\d+\.\d+\.\d+')
```
stavros-k, line 9:
```suggestion
RE_STABLE_VERSION = re.compile(r'v[0-9]+\.[0-9]+\.[0-9]+')
```

stavros-k (changes requested):
Let's revert regex changes.

But indeed the initial issue was that json was missing.
But also that it wanted to strip `v`

Thanks

library/ix-dev/community/adguard-home/upgrade_strategy
xxyzz: Could you explain why `[0-9]` should be used instead of `\d`? And also the above `if not version` line, why `if version is None` shouldn't be used?

xxyzz:
How could I test the changes on my local machine to see whether it works? I briefly looked the `catalog_update` and `catalog_validation` code and the main scripts of both project seem to run in CI and update all the docker images.

library/ix-dev/community/adguard-home/upgrade_strategy
stavros-k: Mostly for consistency with the other apps.
And also I don't know if the specific regex language has support for regex class
I'll let @Qubad786 answer this (and the `if not version` vs `if version is None`) as python is he's area of expertise!

stavros-k (approved):

stavros-k (approved):

library/ix-dev/community/adguard-home/upgrade_strategy
Qubad786: @xxyzz it is fine to use `\d` instead as it's the same thing.
About `version is None` - if you are explicitly looking out for `None` then it is indeed better to check for that explicitly instead of a boolean check i.e `not version`

Qubad786 (approved):

PR #1245, opened 2023-06-06
NAS-122335 / 23.10 / Add `jenkins` to `community` train

stavros-k:
- Runs rootless (UID not configurable)
- Can use certificate for HTTPS
- Additional storage/volumes
- Additional env variables
- Additional Java Jenkins Options / Jenkins Options.
- Validation for duplicate entries / reserved options
- chown on mounted volumes only if parent dir has a mismatch on user/group


library/ix-dev/community/jenkins/templates/_validation.tpl
```diff
@@ -0,0 +1,44 @@
+{{- define "jenkins.validation" -}}
+  {{- if and (not .Values.jenkinsNetwork.http) (not .Values.jenkinsNetwork.https) -}}
+    {{- fail "Jenkins - At least one of [HTTP, HTTPS] must be enabled." -}}
```
Qubad786, line 3: Can both be enabled at the same time?

Qubad786: Just to be sure both http and https work when both options are checked?

stavros-k: Yes, that's correct

Qubad786: In practice it is better to not allow http access when https is enabled, is there specific motivation for this setup?

stavros-k: I agree, only reason I've "allowed" it is if there are any other services that need to interact with Jenkins and do not like https.

Qubad786: Are there any such services at the moment?

stavros-k: I'm not aware of any.

Qubad786: @stavros-k let's remove this behaviour in that case then ;)

Qubad786 (approved):

Issue #1222, opened 2023-05-29
Community Chart Suggestion: Grafana


cykedev:
Grafana

PR #1252, opened 2023-06-08
NAS-122388 / 23.10 / Adds `grafana` to `community` apps

stavros-k:
Adds Grafana.
User can:
- Define additional variables
- Define additional storages
- Pick a certificate and enable HTTPS
- Define plugins to auto install at startup

App runs rootless, uid/gid is configurable
Chown will only happen if directories have a missmatch with either uid or gid


kmoore134:
This solves #1222

library/ix-dev/community/grafana/templates/_grafana.tpl
```diff
@@ -0,0 +1,124 @@
...
+              port: {{ .Values.grafanaNetwork.webPort }}
+              path: /api/health
+            startup:
+              enabled: true
+              type: {{ $scheme }}
+              port: {{ .Values.grafanaNetwork.webPort }}
+              path: /api/health
+      initContainers:
+      {{- include "ix.v1.common.app.permissions" (dict "containerName" "01-permissions"
+                                                        "UID" .Values.grafanaRunAs.user
```
Qubad786, line 49: The permission container will run as the grafana user?

stavros-k: Nop, it runs as root (hardcoded), this value here is used in the script that fixes the perms/ownership

Qubad786 (approved):

PR #1292, opened 2023-06-23
Update questions.yaml

cmillsa2:
Replaced semicolons with colons in URL for server and redirect urls


stavros-k:
Can you please update the `questions.yaml` under `library/ix-dev/enterprise/minio` instead?
While at it, please bump `version` in `Chart.yaml` by `0.0.1` 
Thanks

PR #1309, opened 2023-06-28
NAS-122692 / 23.10 / Fixes a few broken links in "sources"

stavros-k:


stavros-k:
Failed tests for Zerotier and Tailscale are expected.

library/ix-dev/charts/elastic-search/Chart.yaml
```diff
@@ -15,12 +14,12 @@ maintainers:
...
   - name: common
     repository: file://../../../common
-    version: 1.0.8
+    version: 1.0.9
 home: https://www.elastic.co
 icon: https://images.contentstack.io/v3/assets/bltefdd0b53724fa2ce/blt280217a63b82a734/6202d3378b1f312528798412/elastic-logo.svg
 sources:
   - https://www.elastic.co/
-  - https://github.com/truenas/charts/tree/master/charts/elastic-search
+  - https://github.com/truenas/charts/tree/master/library/ix-dev/charts/elastic-search
```
Qubad786, line 22: Why are we changing this path ?

Qubad786 (commented):
Can we please keep logical changes apart i.e PR is supposed to be scoped at fixing source links so we can do that and not touch common in the process and do that in another PR..

library/ix-dev/charts/elastic-search/Chart.yaml
stavros-k: Because this will always show the readme.
If we leave the URL pointing to the catalog, in order to show the readme directly we will have to update the chart version in the link each time

stavros-k:
> Can we please keep logical changes apart i.e PR is supposed to be scoped at fixing source links so we can do that and not touch common in the process and do that in another PR..

Without the common bump it will fail all tests, as the pre-bump version cannot be resolved anymore

stavros-k:
> > Can we please keep logical changes apart i.e PR is supposed to be scoped at fixing source links so we can do that and not touch common in the process and do that in another PR..
> 
> Without the common bump it will fail all tests, as the pre-bump version cannot be resolved anymore

@Qubad786  Removed common bump and after this is merged, I'll follow up with a common update

stavros-k:
Depends on https://github.com/truenas/charts/pull/1326

library/ix-dev/charts/elastic-search/Chart.yaml
Qubad786: I wonder if we should publish latest version readme under `elastic-search` directory directly instead?
The path currently being used for development will be changed in the future as this is for library charts essentially..
What do you think about above proposal instead?

stavros-k: Sounds good

stavros-k: @Qubad786 Fixed some links the other way around in this case.

yocalebo (approved):

PR #1312, opened 2023-06-28
NAS-122694 / 23.10 / Add some validation in webdav and other apps

stavros-k:
- Adds validation for CPU/Memory limits in questions.yaml (there already in the common side)
- Adds validation on "Share name" in webdav
- Makes share list in webdav required
- Adds validation in webdav to make sure only unique share names are defined

~~(Tests failing because apps need to be bumped, will do so after #1309 is merged)~~


stavros-k:
Depends on Depends on https://github.com/truenas/charts/pull/1326

Qubad786 (approved):

PR #1339, opened 2023-07-07
NAS-122858 / 23.10 / Add support for additional library mounts in immich

stavros-k:
Only hostPath type is used here, as the hostPath and mountPath must be exactly the same.

https://immich.app/docs/features/read-only-gallery

### DIFF
--- a/library/ix-dev/community/immich/questions.yaml
+++ b/library/ix-dev/community/immich/questions.yaml
@@ -291,6 +291,28 @@ questions:
                   show_if: [["type", "=", "hostPath"]]
                   immutable: true
                   required: true
+        - variable: additionalLibraries
+          label: Additional Libraries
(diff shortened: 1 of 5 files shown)


library/ix-dev/community/immich/questions.yaml
```diff
@@ -291,6 +291,28 @@ questions:
...
+                label: Storage Entry
+                schema:
+                  type: dict
+                  attrs:
+                    - variable: hostPath
+                      label: Host Path
+                      description: The host path to use for storage.
+                      schema:
+                        type: hostpath
+                        show_if: [["type", "=", "hostPath"]]
```
stavros-k, line 314:
```suggestion
```

Qubad786 (approved):

PR #1343, opened 2023-07-08
Bump immich app version to 1.66.1

xxyzz:
Thanks for adding immich! But the app version is inconsistent with the docker image tag so I'm trying to fix that in this pull request.


stavros-k (approved):
Thanks, good catch

PR #1347, opened 2023-07-10
NAS-122894 / 23.10 / Fix tailscale upgrade strategy

### LINKED ISSUE #1288: Tailscale update from 1.42 to 1.44
Hello,

I would like to know if there is a workflow to ask for update tailscale application from 1.42.0 to 1.44.0 version, please.
Thx for your feedback

Regards

Qubad786:
## Problem

Tail scale app was not upgrading which resulted in users being left behind on newer versions and the reason behind it was that the upgrade strategy was not executable.

## Solution

Make the upgrade strategy executable so next time the cronjob for updating app runs - it will automatically update tailscale app to newest version.

Close #1288


themylogin (commented):
How can we preview bugs like this in the future? File permissions are not the easiest thing to review. Why do we require `update_strategy` to be executable in the first place?

sonicaj:
@themylogin it should be executable so `truenas/catalog_update` can execute it - it can be a python file or any executable as long as it adheres to the input/output expected from it.
@Qubad786 we should have validation similar to https://github.com/truenas/catalog_validation/pull/68 so these cases can be caught earlier in the development lifecycle

sonicaj (approved):

Issue #1369, opened 2023-07-14
WG-Easy App: Fix Hardcoded Device Name


TCB13:
Hello,

There as a fix for the hardcoded `eth0` interface that doesn't work in TrueNAS merged here https://github.com/wg-easy/wg-easy/pull/500

The current version doesn't work because it tries to configure WG to use the `eth0` interface and in modern systems interface names are something like `enp2s0`. The enviromental variable `WG_DEVICE` was introduced to deal with this issue.

Can this be updated in this repo as well? Thank you.

TCB13:
Here is what's happening inside the container:

`````
bash-5.0# ip a
(...)
2: enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff
    inet 172.21.1.205/24 brd 172.21.1.255 scope global dynamic enp2s0
       valid_lft 82313sec preferred_lft 82313sec
(...)
12: wg0: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1420 qdisc noqueue state UNKNOWN group default qlen 1000
    link/none
    inet 10.8.0.1/24 scope global wg0
       valid_lft forever preferred_lft forever


bash-5.0# cat config.js
'use strict';

const { release } = require('./package.json');

module.exports.RELEASE = release;
module.exports.PORT = process.env.PORT || 51821;
module.exports.PASSWORD = process.env.PASSWORD;
module.exports.WG_PATH = process.env.WG_PATH || '/etc/wireguard/';
module.exports.WG_HOST = process.env.WG_HOST;
module.exports.WG_PORT = process.env.WG_PORT || 51820;
module.exports.WG_MTU = process.env.WG_MTU || null;
module.exports.WG_PERSISTENT_KEEPALIVE = process.env.WG_PERSISTENT_KEEPALIVE || 0;
module.exports.WG_DEFAULT_ADDRESS = process.env.WG_DEFAULT_ADDRESS || '10.8.0.x';
module.exports.WG_DEFAULT_DNS = typeof process.env.WG_DEFAULT_DNS === 'string'
  ? process.env.WG_DEFAULT_DNS
  : '1.1.1.1';
module.exports.WG_ALLOWED_IPS = process.env.WG_ALLOWED_IPS || '0.0.0.0/0, ::/0';

[...]

PR #1370, opened 2023-07-14
WG-Easy: Fix Add Variable for Listen Interface/Device

TCB13:
This is an initial PR to fix https://github.com/truenas/charts/issues/1369. It adds a variable on the interface so the user can choose in what interface WG should listen.

It would be very nice if someone with more experience in this project could "chip in" and help make this fix a reality. I'm not sure how to actually fix things futher than this.

Thank you.

### DIFF
--- a/charts/wg-easy/1.0.10/ci/test-values.yaml
+++ b/charts/wg-easy/1.0.10/ci/test-values.yaml
@@ -12,6 +12,7 @@ wgUDPPort: 30290
 webUIPort: 30921
 wgeasy:
   host: wg.domain.com
+  netinterf: en0
   password: secret
   client_mtu: 1420
   keep_alive: 5
--- a/charts/wg-easy/1.0.10/ci/test-values.yaml
+++ b/charts/wg-easy/1.0.10/ci/test-values.yaml
@@ -12,7 +12,6 @@ wgUDPPort: 30290
 webUIPort: 30921
 wgeasy:
   host: wg.domain.com
-  netinterf: en0
   password: secret
   client_mtu: 1420
   keep_alive: 5
--- a/charts/wg-easy/1.0.10/questions.yaml
+++ b/charts/wg-easy/1.0.10/questions.yaml
@@ -88,6 +88,13 @@ questions:
             required: true
             $ref:
               - "definitions/nodeIP"
+        - variable: netinterf
+          label: Listen Interface (Device)
+          description: Set the interface where WireGuard sould listen for incoming connections.
+          schema:
+            type: string
+            required: true
+            default: "en0"
         - variable: password
           label: Password for WebUI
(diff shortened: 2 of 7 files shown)


stavros-k (changes requested):
Hello, please make these changes in `library/ix-dev/charts/wg-easy` directory. 
Thanks

TCB13:
@stavros-k all done :)

TCB13:
> This only affects when hostNetwork is enabled, right? Otherwise it would use the interface from the container. (Which is `eth0`).
> 
> I think we should default this to `eth0` (As the upstream has it). And probably do not set the env var at all when hostNetwork is off, because changing it in this case it will "break" it


As you can see at https://github.com/wg-easy/wg-easy/blob/master/src/config.js#L9 , https://github.com/wg-easy/wg-easy/blob/master/src/config.js#L22 and at https://github.com/wg-easy/wg-easy/blob/master/README.md?plain=1#L86 the upstream replaced the static `eth0` with the `WG_DEVICE` env variable. It also fallbacks to `eth0` so nothing will break. In my commit I also set `eth0` as default value for the variable.

Either way, the current solution at truenas is broken with hostNetwork is enabled and disabling results in additional and unnecessary CPU load.

stavros-k:
> > This only affects when hostNetwork is enabled, right? Otherwise it would use the interface from the container. (Which is `eth0`).
> > I think we should default this to `eth0` (As the upstream has it). And probably do not set the env var at all when hostNetwork is off, because changing it in this case it will "break" it
> 
> As you can see at https://github.com/wg-easy/wg-easy/blob/master/src/config.js#L9 , https://github.com/wg-easy/wg-easy/blob/master/src/config.js#L22 and at https://github.com/wg-easy/wg-easy/blob/master/README.md?plain=1#L86 the upstream replaced the static `eth0` with the `WG_DEVICE` env variable. It also fallbacks to `eth0` so nothing will break. In my commit I also set `eth0` as default value for the variable.
> 
> Either way, the current solution at truenas is broken with hostNetwork is enabled and disabling results in additional and unnecessary CPU load.

Okay, but we'll have to wait until a new container is release with this change first.

TCB13:
Lets leave this open then. :)

stavros-k:
Considering the upcoming migration #1403, I'll close this one. This is because it will end up with a lot of conflicts anyway.

Feel free to open a new PR after the linked migration is merged.
Otherwise I'll handle it once upstream releases a new image

Thanks

TCB13:
> Feel free to open a new PR after the linked migration is merged.
> Otherwise I'll handle it once upstream releases a new image

I ended up abandoning TrueNAS altogether as most images on the "charts" are broken in one or more ways and the base system (TrueNAS Scale) doesn't seem mature enough for any kind of production usage.

Thank you anyways.

PR #1372, opened 2023-07-15
Update immich v1.67.2

xxyzz:


stavros-k:
Hello, this have been updated in #1371 
Thanks

Issue #1305, opened 2023-06-28
[Issue] NextCloud Application - Web Server not configured to resolve carddav/caldav


T13nou:
Hi !

This issue report is tied to discussion on TN Forum : https://www.truenas.com/community/threads/nextcloud-moved-from-truecharts-back-to-official.110820/

**Context**
TrueNAS-SCALE-22.12.3.1
NextCloud Official App 27.0.0_1.6.31
Catalog:TRUENAS
Train:charts

**Issue Report**
Issue is displayed in NextCloud in "Security & Configuration warnings section"

> * Your web server is not properly set up to resolve "/.well-known/caldav". Further information can be found in the [documentation]
> * Your web server is not properly set up to resolve "/.well-known/carddav". Further information can be found in the [documentation]

**Possible fix**
Don't know

Issue #1302, opened 2023-06-28
nextcloud  upload 413 Request Entity Too Large


kailin-hub:
NextCloud upload reported an error
`Error: [EFAULT] < ?xml version="1.0" encoding="utf-8"? > Sabre\DAV\Exception\BadRequest 预期文件大小为 10664406952字节，实际从 Nextcloud 客户端读入并写入 Nextcloud 存储空间的大小为 0字节。可能是发送端发生了网络问题，或者是服务器写入存储设备时发生错误。 : 413 Request Entity Too Large`
this is nextcloud.log:
`[no app in context] 错误: Sabre\DAV\Exception\BadRequest: 预期文件大小为 6507042880字节，实际从 Nextcloud 客户端读入并写入 Nextcloud 存储空间的大小为 0字节。可能是发送端发生了网络问题，或者是服务器写入存储设备时发生错误。 at <<closure>>

0. /var/www/html/apps/dav/lib/Connector/Sabre/Directory.php line 149
   OCA\DAV\Connector\Sabre\File->put("*** sensitive parameters replaced ***")
1. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 1098
   OCA\DAV\Connector\Sabre\Directory->createFile("*** sensitive parameters replaced ***")
2. /var/www/html/3rdparty/sabre/dav/lib/DAV/CorePlugin.php line 504
   Sabre\DAV\Server->createFile("*** sensitive parameters replaced ***")
3. /var/www/html/3rdparty/sabre/event/lib/WildcardEmitterTrait.php line 89
   Sabre\DAV\CorePlugin->httpPut(["Sabre\\HTTP\\Request"], ["Sabre\\HTTP\\Response"])
4. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 472
   Sabre\DAV\Server->emit("method:PUT", [["Sabre\\HTTP\\ ... ]])
5. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 253
   Sabre\DAV\Server->invokeMethod(["Sabre\\HTTP\\Request"], ["Sabre\\HTTP\\Response"])
6. /var/www/html/3rdparty/sabre/dav/lib/DAV/Server.php line 321
   Sabre\DAV\Server->start()
7. /var/www/html/apps/dav/lib/Server.php line 364
   Sabre\DAV\Server->exec()
8. /var/www/html/apps/dav/appinfo/v2/remote.php line 35
[...]

PR #1380, opened 2023-07-18
NAS-122852 / 23.10 / Allow setting timeouts for nginx and upload size to both php and nginx

stavros-k:
Regarding #1305

When a certificate is configured, we spin up nginx, which handles the redirects.
Nginx was listening internally on 443, which made nginx to use that port for redirects too.
I've adjusted nginx config to always append the port and also use the user defined port.

This should not cause issues to any setups. **Unless** there is someone that have **another** reverse proxy infront of nginx.
In such case, assuming the "other" reverse proxy uses some other port, the redirect URLs won't match the correct port.

---

Regarding #1302

The max size and timeouts in nginx configuration was hardcoded, I've added 2 fields to allow users define a different value.
Those fields only accept `int` and not the **unit**. This is to avoid typo's and misconfiguration.

The max size field is always visible, as it sets both `client_max_body_sized` and `PHP_UPLOAD_LIMIT`
The timeouts field is only shown when cert is configured (Only then we spawn nginx)


library/ix-dev/charts/nextcloud/questions.yaml
```diff
@@ -116,6 +136,16 @@ questions:
...
             type: boolean
             default: false
+        - variable: max_upload_size
+          label: "Max Upload Size (Giga Bytes)"
+          description: |
+            Applies the timeout to the following settings:</br>
+            - client_max_body_size in nginx</br>
+            - post_max_size and upload_max_filesize in php</br>
+          schema:
+            type: int
```
Qubad786, line 146: Should we have a minimum here and above?

stavros-k: Agreed

Qubad786 (approved):

645340633:
I use homeaassistant addon ( [hassio-nextcloud-backup](https://github.com/Sebclem/hassio-nextcloud-backup)) to backup the snapshot  files, but occurs Fail to upload snapshot files to nextcloud (HTTPError: Response code 413 (Request Entity Too Large)) ! 
issue What should I do to solve the problem

My snapshot backup file is 6GB ,This is what I set，but it seems doesn't work

PR #1403, opened 2023-07-25
NAS-123206 / 23.10 / migrate `wg-easy` to the new common library

stavros-k:
Tested upgrades from 1.0.11/1.0.12 to 2.0.0.
Everything worked fine on my side.

(On .11 -> .12 only some links for images changed)

During the execution of `migrations` scripts, the `ixChartMetadata.upgradeMetadata`, is not populated,
So we migrate values there, check on `helm` side if the versions are ok to proceed, if not fail.

Checked and the old values are restored in that case.


sonicaj (commented):
I see various issues with this, i think it's better to have a huddle and discuss this a bit more. Let's sync on slack

stavros-k:
> I see various issues with this, i think it's better to have a huddle and discuss this a bit more. Let's sync on slack

Simplified the `migrate` script a lot, and moved some checks on helm side due to upgradeMetadata not being available at the time of migration.

sonicaj (approved):

PR #1412, opened 2023-07-27
NAS-123258 / 23.10 / allow to expose daemon port

### LINKED ISSUE #1411: Deluge app should open daemon port for thin clients
The Deluge app in truenas is 95% of the way there but does not expose the daemon port, only the 'torrent' and 'web' ports, it's defined as exposed in the original linuxcontainers dockerfile however https://github.com/linuxserver/docker-deluge/blob/master/Dockerfile#L58C13-L58C18 

the default daemon port is 58846 and if it's exposed then thin clients like deluge-gtk can connect and manage the server over that connection

stavros-k:
Fixes #1411

### DIFF
--- a/library/ix-dev/community/deluge/ci/basic-values.yaml
+++ b/library/ix-dev/community/deluge/ci/basic-values.yaml
@@ -8,3 +8,4 @@ delugeStorage:
 
 delugeNetwork:
   exposeDaemon: true
+  daemonPort: 32100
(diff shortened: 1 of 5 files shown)


library/ix-dev/community/deluge/ci/basic-values.yaml
```diff
@@ -5,3 +5,6 @@ delugeStorage:
   downloads:
     type: hostPath
     hostPath: /mnt/{{ .Release.Name }}/downloads
+
+delugeNetwork:
+  exposeDaemon: true
```
stavros-k, line 10:
```suggestion
  exposeDaemon: true
  daemonPort: 32100
```

Qubad786 (approved):

PR #1413, opened 2023-07-27
NAS-123265 / 23.10 / adds `tinyMediaManager` to `community` train

stavros-k:


library/ix-dev/community/tiny-media-manager/upgrade_strategy
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/python3
+import json
+import re
+import sys
+
+from catalog_update.upgrade_strategy import semantic_versioning
+
+
+RE_STABLE_VERSION = re.compile(r'\d+\.\d+\.\d+(\.\d+)?')
```
Qubad786, line 9: So it can be some thing like `1.1.12.13`?

Qubad786 (approved):

library/ix-dev/community/tiny-media-manager/upgrade_strategy
stavros-k: Yes, its varying between 3 and 4 "parts"
image

PR #1420, opened 2023-07-31
nextcloud: allow tuning opcache memory consuption

stavros-k:
Fixes #1418


library/ix-dev/charts/nextcloud/questions.yaml
```diff
@@ -148,6 +148,14 @@ questions:
             default: 3
             min: 1
             required: true
+        - variable: opcache_memory_consumption
+          label: "Opcache Memory Consumption (Mega Bytes)"
+          description: "Configures the memory consumption of the opcache"
+          schema:
+            type: int
+            min: 128
```
Qubad786, line 156: Should there be a max?

Qubad786 (approved):

library/ix-dev/charts/nextcloud/questions.yaml
stavros-k: Lets set it to 1024? I think it should be more than enough?

Qubad786: I think yes

PR #1465, opened 2023-08-21
NAS-123507 / 24.04 / fix URL generation for nextcloud when SSL is enabled

stavros-k:
When certificate is selected and nginx is deployed, nginx will rewrite the URL, adding the port (9001 default).

But in cases a user have a firewall rule (outside of TrueNAS) directing `443` traffic to `9001` port, the rewrite should not include the port.

This PR adds a flag that allows the user to decide if the port (9001) should be included in the url rewrite, or use the port 443 (which in this case the port does not need to be appended at all).

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html#nginx


library/ix-dev/charts/nextcloud/questions.yaml
```diff
@@ -86,6 +86,14 @@ questions:
             min: 30
             default: 60
             required: true
+        - variable: use443
+          label: "Use 443 for URL rewrites"
```
Qubad786, line 90: Maybe add `port` here as well?

Qubad786 (approved):

PR #1477, opened 2023-08-25
NAS-123719 / 24.04 / Fix deluge torrent port

### LINKED ISSUE #1474: Deluge app can never accept incoming connections
The deluge chart as it is currently setup enforces that the deluge program listen on port 6881 for it's "torrent port", allowing the exposed port to be configured to whatever high port the user wants. 

This presents a problem as deluge will announce itself as being available on <some ip> on port 6881, which is never connectable as the actual exposed port is somewhere higher. port 6881 itself isn't allowed as an exposed port in truenas so we can't solve the problem that way. 

The fix would be to allow the port deluge knows about as well as the exposed port to be set to the same value, maybe just allowing any port to be exposed for this purpose?

stavros-k:
- Splits some code into multiple files for readability
- Adjusts internal (container's) torrent port based on the user defined values (Fixes #1474)


library/ix-dev/community/deluge/templates/_deluge.tpl
```diff
@@ -48,79 +48,39 @@ workload:
...
+          type: init
+          imageSelector: image
+          securityContext:
+            runAsUser: 0
+            runAsGroup: 0
+            runAsNonRoot: false
+            readOnlyRootFilesystem: false
+            capabilities:
+              add:
+                - CHOWN
```
Qubad786, line 63: Do we need to update our static record where we maintain these capabilities?

stavros-k: We have them listed already for the main container, which is the same container.

Qubad786 (approved):

PR #1516, opened 2023-09-06
NAS-123941 / 24.04 / Vaultwarden - Add TZ config option

### LINKED ISSUE #1515: VAULTWARDEN time zone
Could be possible to implement the possibility to choose the time zone?
Impossible to use TZ in an Additional Environment Variables because is already in use. With the default time zone is impossible to use 2FA app.

stavros-k:
Fixes #1515


library/ix-dev/community/vaultwarden/questions.yaml
```diff
@@ -29,6 +29,16 @@ portals:
     path: "$kubernetes-resource_configmap_portal_admin_path"
 
 questions:
+  - variable: TZ
```
Qubad786, line 32: To be sure - this will automatically get picked up by common so it reflects this in the container?

Qubad786 (approved):

CirKanus:
Despite the TZ variable, vaultwarden steal think I am in UTC, but I don’t.

stavros-k:
If you shell in the container, and type `env | grep TZ` does it reflect your selection?
If it does, you have to reach out to vaultwarden support channels and see why it's not picking it up

CirKanus:
> If you shell in the container, and type `env | grep TZ` does it reflect your selection? If it does, you have to reach out to vaultwarden support channels and see why it's not picking it up

It does.

stavros-k:
> > If you shell in the container, and type `env | grep TZ` does it reflect your selection? If it does, you have to reach out to vaultwarden support channels and see why it's not picking it up
> 
> It does.

Thanks for confirming, in that case you have to reach out to vaultwarden developers and see why it does not respect the variable

CirKanus:
In me case the problem where in the hardware clock, thanks for the help.

Issue #1452, opened 2023-08-14
Nextcloud data volume unexpected behavior


kridgo:
I am migrating from TrueCharts Nexctloud application to the TrueNAS one due to the broken app update situation in TrueCharts.
When setting up the app, I specified a host path for the Nextcloud Data Volume.

The content of the host path then looks as follows:
```
drwxr-xr-x  9 root     root      9 Aug 14 21:48 .
drwxr-xr-x  4 root     root      4 Aug 14 21:46 ..
drwxr-xr-x  2 www-data www-data 15 Aug 14 21:48 config
drwxr-xr-x  2 www-data www-data  2 Aug 14 21:48 custom_apps
drwxr-xr-x  2 www-data www-data  5 Aug 14 21:48 data
drwxr-xr-x 15 www-data www-data 30 Aug 14 21:48 html
drwxr-xr-x  4 root     root      4 Aug 14 21:48 root
drwxr-xr-x  3 www-data www-data  4 Aug 14 21:48 themes
drwxr-xr-x  2 root     root      2 Aug 14 21:48 tmp
```
The content of `html` looks like this:
```
drwxr-xr-x 15 www-data www-data    30 Aug 14 21:48 .
drwxr-xr-x  9 root     root         9 Aug 14 21:48 ..
-rw-r--r--  1 www-data www-data  3127 Aug 14 21:48 .htaccess
-rw-r--r--  1 www-data www-data   101 Aug 14 21:48 .user.ini
drwxr-xr-x 45 www-data www-data    52 Aug 14 21:48 3rdparty
-rw-r--r--  1 www-data www-data 23796 Aug 14 21:48 AUTHORS
-rw-r--r--  1 www-data www-data 34520 Aug 14 21:48 COPYING
drwxr-xr-x 50 www-data www-data    50 Aug 14 21:48 apps
drwxr-xr-x  2 root     root         2 Aug 14 21:48 config
-rw-r--r--  1 www-data www-data  4095 Aug 14 21:48 console.php
drwxr-xr-x 24 www-data www-data    31 Aug 14 21:48 core
-rw-r--r--  1 www-data www-data  6317 Aug 14 21:48 cron.php
drwxr-xr-x  2 root     root         2 Aug 14 21:48 custom_apps
drwxr-xr-x  2 root     root         2 Aug 14 21:48 data
drwxr-xr-x  2 www-data www-data   172 Aug 14 21:48 dist
[...]

PR #1557, opened 2023-09-21
add additional mount for user data

kridgo:
As described in  #1452, currently you can not mount the nextcloud user data to a different dataset than the other nextcloud data (content from `/var/www/html`).
You are able to change the user data directory inside the nextcloud container, however the host path volume configuration will put it into one directory together with the `html` folder on the host.

This PR adds another volume for the user data to provide the user with the possibility to separate user data from the `html`-folder.

### DIFF
--- a/charts/nextcloud/1.6.48/questions.yaml
+++ b/charts/nextcloud/1.6.48/questions.yaml
@@ -137,7 +137,7 @@ questions:
             required: true
         - variable: datadir
           label: "Nextcloud data directory"
-          description: "Configures the data directory where nextcloud stores all files from the users"
+          description: "Configures the data directory where nextcloud stores all files from the users (inside the container). When changing this, make sure to specify a host path for Nextcloud User Data Volume in the section storage."
           schema:
             type: path
             default: "/var/www/html/data"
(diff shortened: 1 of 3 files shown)


charts/nextcloud/1.6.48/questions.yaml
```diff
@@ -285,6 +285,43 @@ questions:
                       schema:
                         type: hostpath
                         required: true
+        - variable: nextcloud-userdata
+          label: "Nextcloud User Data Volume"
+          schema:
+            type: dict
+            show_if: [[ "nextcloud.datadir", "!=", "/var/www/html/data" ]]
```
kridgo, line 292: When the user data remains in the folder `/var/www/html/data`, the additional user data volume should not be created, to support the user hints were added in the variable descriptions. I also wanted to hide the new `Nextcloud User Data Volume` when the default value was not changed by the user. However, the `show_if` directive seems to be misconfigured or being ignored.

stavros-k:
To be honest, I'm not a fan of this, its too "hacky". Which will only make it harder for future changes/additions.

I can see how separate volumes can be beneficial, 
this would need more thought and most likely a migration to accomplish.

kridgo:
I see your point regarding the "hackiness" of the PR.
At least for me however, the app is not usable right now, as I need to separate userdata from the html-folder. I am wondering how other users of the chart deal with this.

Do you have a suggestion how to implement the separation in a "cleaner" way?

For me it would be fine if I could specify the userdata folder somewhere in the container and then mount this via an extra host path volume. But it will not let me mount the path as another volume due to folder already being mounted as sub path of the data mount:
https://github.com/truenas/charts/blob/89198f9b80c6403cfc6145404b58d6934ac6cee1/charts/nextcloud/1.6.49/templates/deployment.yaml#L173

stavros-k:
> Do you have a suggestion how to implement the separation in a "cleaner" way?

Ideally all "subfolders" would be on a different volume.
For users using hostPath it should be very easy to migrate (using the migrations directory, automatically).
I need to think how the ixVolumes would be migrated.

stavros-k:
Closing this one now, should be a much better situation once #2377 lands.
Sorry it took that long.
Thanks

Issue #1277, opened 2023-06-19
Community Chart Suggestion: 2FAuth


ghost:
https://docs.2fauth.app/

PR #1571, opened 2023-09-25
Add `2FAuth` to `community` train

stavros-k:
Closes https://github.com/truenas/charts/issues/1277


library/ix-dev/community/two-f-auth/Chart.yaml
```diff
@@ -0,0 +1,27 @@
+name: two-f-auth
```
Qubad786, line 1: Should we name it something different?

stavros-k: Would two-fauth be okay ? or just one word twofauth?
We can't start with a digit

Qubad786: `twofactor-auth` perhaps?

stavros-k: I'm fine with that, yea

Qubad786 (approved):

PR #1574, opened 2023-09-26
Latest version of Plex app

mosaati:
This is the correct latest version as per https://hub.docker.com/r/plexinc/pms-docker/tags

### DIFF
--- a/charts/plex/1.7.55/ix_values.yaml
+++ b/charts/plex/1.7.55/ix_values.yaml
@@ -1,7 +1,7 @@
 image:
   pullPolicy: IfNotPresent
   repository: plexinc/pms-docker
-  tag: 1.32.6.7468-07e0d4a7e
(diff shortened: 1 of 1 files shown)


stavros-k (changes requested):
Changes should be done only under `library/ix-dev/charts/plex`

Other directories are overwritten on each push.

Thanks,

Beyond that, what happens to users that have already upgraded? Will plex work if the user downgrades?

mosaati:
> Changes should be done only under `library/ix-dev/charts/plex`
> 
> Other directories are overwritten on each push.
> 
> Thanks,
> 
> Beyond that, what happens to users that have already upgraded? Will plex work if the user downgrades?

It is a rollback in the release tag version but the app itself internally is an upgrade from "1.32.5.7349" to "1.32.5.7516".

mosaati:
I will create another pull request under library/ix-dev/charts/plex

PR #1580, opened 2023-09-27
Plex_Rollback

mosaati:
This change is for Plex latest version that is a rollback in the container version number from [1.32.6.7468-07e0d4a7e] to [1.32.5.7516-8f4248874]. However the app itself is an upgrade from 1.32.5.7349 to 1.32.5.7516.

https://hub.docker.com/r/plexinc/pms-docker/tags
270867445-aa19d1cb-c942-4698-b695-d66ff758c07e


stavros-k (changes requested):
Please update versions in Chart.yaml
Thanks

Issue #1112, opened 2023-04-12
Hard to pull images on ghcr.io & Cannot update the storage configuration after installing


BreakVoid:
After I install the new community version of qBittorrent, I found two issues.

1. The previous official chart of qBittorrent uses the images from linuxserver(lscr.io/linuxserver/qbittorrent:<version>), while the current community version uses the images from ghcr.io, which is hard to pull in China. I have to manually pull it by setting a proxy. Could it use the previous image source?

2. The storage configuration cannot be modified after installation (as shown in the figure).
<img width="744" alt="截屏2023-04-12 11 50 23" src="https://user-images.githubusercontent.com/8384350/231345070-101734b6-e01d-43c7-a430-555eb894242c.png">

My TrueNAS version: 22.12.2

alexislefebvre:
Do you have the same issues when you download images from Red Hat's [Quay Container Registry](https://quay.io/)? It may be used as an alternative to GHCR, if it's possible to build images once then push them to GHCR and Quay. Then TrueNAS would have to allow to choose between GHCR and Quay as the source of the images.

Example:

```bash
docker pull quay.io/prometheus/prometheus
```

BreakVoid:
The following command is executed successfully on my NAS.
```docker pull quay.io/prometheus/prometheus```
I think it should be a good solution if there is no other issue.

stavros-k:
A better improvement here would be to "Support at least 2 registries when possible"

stavros-k:
Unfortunately there are currently no plans to support multiple registries per image as it's not always in sync and hard to maintain.

But you can now use docker/docker compose in 24.10 and later, and use any image you want.
Also the ability to convert apps to custom yaml apps is coming on next release https://github.com/truenas/webui/pull/11109
So it should help as well

PR #1604, opened 2023-10-05
Use official ghcr.io image for community/immich

jbgosselin:
Hi !

I'm quite often hit by docker-hub's rate limiting when updating images, especially for immich.

After some investigation, I found that immich is offering their images via ghcr.io Github's Registry.

It seems that it is also what is used in their official docker-compose file https://github.com/immich-app/immich/blob/main/docker/docker-compose.yml

Additionally, the image in docker-hub (under `altran1502`) has the same hash than the one on ghcr.io, so I guess it's just pushed on both registries.

### DIFF
--- a/library/ix-dev/community/immich/Chart.yaml
+++ b/library/ix-dev/community/immich/Chart.yaml
@@ -3,7 +3,7 @@ description: Immich
 annotations:
   title: Immich
 type: application
-version: 1.0.26
+version: 1.0.27
 apiVersion: v2
 appVersion: 1.81.1
 kubeVersion: '>=1.16.0-0'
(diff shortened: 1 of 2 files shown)


stavros-k:
Hello, this will create issues like this: https://github.com/truenas/charts/issues/1112

jbgosselin:
I also understand that `immich` is in the community repo, so it may only be fore truely supported Apps by Truenas.

stavros-k:
> could it be also an idea for Truenas to own their own registry in the future

Lets close this one and follow this issue regarding how the rate limit issue will be handled
https://github.com/truenas/charts/issues/1112

jbgosselin:
@stavros-k regarding authentication for docker-hub, I opened a ticket on ix-system atlassian.
With truenas Scale Cobia, I found a nice way to authenticate with docker hub.
https://ixsystems.atlassian.net/browse/NAS-124474?focusedCommentId=222587
If you want to push the idea, feel free !

Issue #1322, opened 2023-07-03
Community Chart Suggestion: listmonk


pabloeisenhut:
https://listmonk.app/

PR #1633, opened 2023-10-11
Add `listmonk` to `community` train

stavros-k:
Closes https://github.com/truenas/charts/issues/1322

---

https://ixsystems.atlassian.net/browse/TNCHARTS-297


library/ix-dev/community/listmonk/templates/_configuration.tpl
```diff
@@ -0,0 +1,41 @@
+{{- define "listmonk.configuration" -}}
+  {{- $fullname := (include "ix.v1.common.lib.chart.names.fullname" $) -}}
+
+  {{- $dbHost := (printf "%s-postgres" $fullname) -}}
```
sonicaj, line 4: Out of curiosity, can the name change ?

sonicaj (approved):

library/ix-dev/community/listmonk/templates/_configuration.tpl
stavros-k: The fullname is based on the name a user gives to the app, so it will be unique across the cluster.
`postgres` part is not changing, hardcoded in the template that generates the postgres pod.

PR #1670, opened 2023-10-25
Adds smb-pv-pvc option to apps

stavros-k:
This PR only adds the smb-pv-pvc option to "Additional Storage" section.

For individual storages (eg config or data) needs investigation per app and will be implemented later, to make sure those are not storing any kind of sqlite based files, as its very likely to not play nice with smb.


sonicaj (approved):
LGTM

PR #1697, opened 2023-11-02
minio/charts: fix wrong order in questions that cause fields to be hidden

stavros-k:
`minioDomain` has a `show_if` that is based on the certificate.
When the variable in the condition is after the condition both fields are hidden in the UI.


Qubad786 (approved):
Can you create a ticket please because this should not matter..so either this is a middleware or a UI issue?

Issue #1669, opened 2023-10-25
Community Chart Suggestion: Mealie


in2erval:
[Mealie](https://nightly.mealie.io/) is a self-hosted recipe manager and meal planner.

Docker image: ghcr.io/mealie-recipes/mealie:nightly

postgres-based installation: https://nightly.mealie.io/documentation/getting-started/installation/postgres/

PR #1701, opened 2023-11-03
Add `mealie` to `community` train

stavros-k:
Closes https://github.com/truenas/charts/issues/1669

---

https://ixsystems.atlassian.net/browse/TNCHARTS-322

Depends on https://github.com/truenas/charts/pull/1702, so it can use the storageOptions template.


library/ix-dev/community/mealie/upgrade_strategy
```diff
@@ -0,0 +1,31 @@
...
+
+from catalog_update.upgrade_strategy import semantic_versioning
+
+
+RE_STABLE_VERSION = re.compile(r'v\d+\.\d+\.\d+')
+
+
+def newer_mapping(image_tags):
+    key = list(image_tags.keys())[0]
+    tags = {t.strip('v'): t for t in image_tags[key] if RE_STABLE_VERSION.fullmatch(t)}
```
Qubad786, line 14: I am doubtful this will work..it will work similarly with `v1.0.0-RC1.1` and `v1.0.0-BETA`

stavros-k: Plan is to only automatically upgrade to the next "stable" release.
RC versions don't have a specific format, so we can't really account for that.
I'm hoping that there won't be another RC release upstream, but if it does we can upgrade that one manual once.

Qubad786 (approved):

PR #1688, opened 2023-10-30
searxng - ACL for hostPath + ixVolume migration

stavros-k:
A migration added under `migrations` dir for existing installations. Also a minor migration added in helm side, but that's only for CI, as we test upgrades with previous chart too. It can be safely removed on the next bump.

---

Related:
https://github.com/truenas/catalog_validation/pull/78

Maybe we should comment out aclEntries from schema for now, until the next Cobia release?
This should allow us to perform the migration on the rest of the apps, and will only need to uncomment it when the changes are in

---

After this is approved and finalized, I'll proceed updating/migrating the rest of the apps

Depends on https://github.com/truenas/charts/pull/1702

PR #1702, opened 2023-11-03
Storage entry template to reduce boilerplate

stavros-k:
Adds a template for the storage entry to use in apps, this will allow to easily add extra fields in the future, with only adapting the questions on the apps

Also bumps CI deps

---

New structure for hostPath and ixVolume

Also readOnly is added as option for all storage types (hostPath, ixVolume, smb).
This option is added only for additionalStorages.

```yaml
appStorage:
  additionalStorages:
   # hosPath ACL true
    - type: hostPath
      readOnly: true
      mountPath: /data1
      hostPathConfig:
        aclEnable: true
        acl:
          path: /mnt/pool/data1
  
    # hosPath ACL false
    - type: hostPath
      readOnly: true
      mountPath: /data2
      hostPathConfig:
        aclEnable: false
        hostPath: /mnt/pool/data2
  
    # ixVol ACL true
    - type: ixVolume
      mountPath: /data3
      ixVolumeConfig:
        datasetName: data3
        aclEnable: true
  
    # ixVol ACL false
    - type: ixVolume
      mountPath: /data4
      ixVolumeConfig:
        datasetName: data4
        aclEnable: false

    # SMB (no acl here)
    - type: smb-pv-pvc
      mountPath: /data5
      smbConfig:
        server: 10.0.0.1
        share: share-name
        domain: domain
        username: username
        password: password
        size: 1
```

### DIFF
--- a/.github/ct-install-config/common-ct-lint.yaml
+++ b/.github/ct-install-config/common-ct-lint.yaml
@@ -1,6 +1,6 @@
 remote: origin
 target-branch: master
-helm-extra-args: --timeout 600s --debug
+helm-extra-args: --debug
 lint-conf: .github/ct-install-config/lint-conf.yaml
 chart-yaml-schema: .github/ct-install-config/chart_schema.yaml
 # Check that the version in Chart.yaml is incremented
--- a/library/common/templates/app_functions/_postgres.tpl
+++ b/library/common/templates/app_functions/_postgres.tpl
@@ -192,9 +192,7 @@ pgBackup (required): Data persistence configuration for backup
 
 postgresdata:
   enabled: true
-  type: {{ $data.type }}
-  datasetName: {{ $data.datasetName | default "" }}
-  hostPath: {{ $data.hostPath | default "" }}
+  {{- include "ix.v1.common.app.storageOptions" (dict "storage" $data) | nindent 2 }}
   targetSelector:
     postgres:
       postgres:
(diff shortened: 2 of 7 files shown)


.github/ct-install-config/common-ct-lint.yaml
```diff
@@ -1,6 +1,6 @@
 remote: origin
 target-branch: master
-helm-extra-args: --timeout 600s --debug
+helm-extra-args: --debug
```
sonicaj, line 3: Curious, not required anymore ?

library/common/templates/app_functions/_postgres.tpl
```diff
@@ -203,9 +201,7 @@ postgresdata:
         mountPath: /mnt/directories/postgres_data
 postgresbackup:
   enabled: true
-  type: {{ $backup.type }}
-  datasetName: {{ $backup.datasetName | default "" }}
-  hostPath: {{ $backup.hostPath | default "" }}
+  {{- include "ix.v1.common.app.storageOptions" (dict "storage" $backup) | nindent 2 }}
```
sonicaj, line 204: Just to be sure, `$backup` when used here will always have relevant data properly adjusted already ?

sonicaj (approved):
Are we actually changing the format of questions atm as well because that will result in a decent number of migrations to be written ?

.github/ct-install-config/common-ct-lint.yaml
stavros-k: Was never valid option, just didnt throw an error before!

library/common/templates/app_functions/_postgres.tpl
stavros-k: Yes! there is a sanity check few lines above to check the dictionary is not empty,
And if there are any required values missing, common chart will throw an error too

stavros-k:
> Are we actually changing the format of questions atm as well because that will result in a decent number of migrations to be written ?

Yes, with the current format, it is impossible to add ACLs.

I've created a mostly standardized migration here: 
https://github.com/truenas/charts/pull/1688/files#diff-5ddac63049175241bcba93821613b955668822fa67738f7ae4257ea1bca09ac0

PR #1705, opened 2023-11-05
Add readOnly option to Immich's additionalLibraries

telendt:
It would be helpful if `additionalLibraries` for Immich app allowed for read-only mounts of host paths. These are frequently used for [External Libraries](https://immich.app/docs/features/libraries#external-libraries)  and mounting them in read-only mode is a good practice.

Screenshot 2023-11-06 at 00 53 37

### DIFF
--- a/community/immich/1.0.33/ci/extra-values.yaml
+++ b/community/immich/1.0.33/ci/extra-values.yaml
@@ -17,6 +17,7 @@ immichStorage:
   additionalLibraries:
     - hostPath: /mnt/{{ .Release.Name }}/additionalLibrary1
     - hostPath: /mnt/{{ .Release.Name }}/additionalLibrary2
+      readOnly: true
   pgData:
     type: hostPath
     hostPath: /mnt/{{ .Release.Name }}/pgData
--- a/community/immich/1.0.33/questions.yaml
+++ b/community/immich/1.0.33/questions.yaml
@@ -295,6 +295,12 @@ questions:
                       schema:
                         type: hostpath
                         required: true
+                    - variable: readOnly
+                      label: "Read Only"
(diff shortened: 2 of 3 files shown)


telendt:
I have no idea how one runs checks for the MR (I probably can't do it myself), but I am reporting that I tested this change on my system (custom catalog created from my fork) and it works.

stavros-k (changes requested):
Changes should be done under `library/ix-dev/community/immich`
The versioned directories are autogenerated from the above path.

But due to a change that is going to happen in the next days, that will add some features (including readOnly), I'll prefer to not do this currently, as it will make migration harder.

If the initial PR #1688 is reviewed/merged, I will continue will all the other apps asap.

telendt:
Ok, so if I get it right there is no point in me fixing it? Will you take over?

I think there are many cases when people would like to share their media files via read-only host paths. They feel more safe this way (even if the app says it doesn't change files in that volume). E.g. there is this discussion about Jellyfin:
https://www.truenas.com/community/threads/how-to-give-jellyfin-read-only-access.113118/

Feel free to close this MR if you plan to work on it already. But also, if you think I can help somehow please let me know.

stavros-k:
> Ok, so if I get it right there is no point in me fixing it? Will you take over?

Correct!

> I think there are many cases when people would like to share their media files via read-only host paths. They feel more safe this way (even if the app says it doesn't change files in that volume). E.g. there is this discussion about Jellyfin: https://www.truenas.com/community/threads/how-to-give-jellyfin-read-only-access.113118/

It will be an option to all apps soon!

> Feel free to close this MR if you plan to work on it already. But also, if you think I can help somehow please let me know.

I'll close this one for now!

Issue #1719, opened 2023-11-09
immich GPU should be optional


hlstwizard:
I'm deploying immich in my NAS based on AMD CPU and it doesn't support hardware acceleration of immich.

I would strongly that make GPU optional to immich.

stavros-k:
https://github.com/truenas/charts/pull/1720#issuecomment-1810324489

PR #1720, opened 2023-11-09
Make GPU in Immich optional

hlstwizard:


hlstwizard:
Issue #1719

stavros-k:
If you don't select a GPU it won't allocate a GPU.
By default is set to 0 GPUs. I'm not sure what this trying to solve?

image

stavros-k:
Closing this, as it is both not providing enough details why this is needed, and also autogenerated files are changed.
Changes should be done only under `library/ix-dev/*` 
Feel free to ping me, with more details why this is needed so we can discuss it.

Thanks

PR #1724, opened 2023-11-10
Update Chart.yaml  Pihole

LierschIT:
Updated Pihole Version to 2023.10.0


stavros-k:
Hello, I'm going to close this as it's being handled in #1726 

Just a note for the future, only do changes under `library/ix-dev/*` directories

PR #1753, opened 2023-11-16
n8n - migrate storage section

stavros-k:
- Updates questions.yaml to reflect the new schema
- Migrates predefined storage sections to the new format.
- Removes `chown` at initial installation


library/ix-dev/community/n8n/migrations/migrate
```diff
@@ -0,0 +1,85 @@
...
+
+def storageMigrate(storage):
+    delKeys = []
+    if storage['type'] == 'hostPath':
+        # Check if the key exists, if not we have already migrated
+        if not storage.get('hostPath'):
+            return storage
+
+        hostPath = storage['hostPath']
+        storage['hostPathConfig'] = { 'hostPath': hostPath }
```
Qubad786, line 15: unnecessary spaces
`{'hostPath': hostPath}`

library/ix-dev/community/n8n/migrations/migrate
```diff
@@ -0,0 +1,85 @@
...
+        delKeys.append('hostPath')
+
+    if storage['type'] == 'ixVolume':
+        # Fixes an invalid key carried from values.yaml
+        if storage.get('hostPath'):
```
Qubad786, line 20: This can be done like this `storage.pop(key, None)`

library/ix-dev/community/n8n/migrations/migrate
```diff
@@ -0,0 +1,85 @@
...
+
+        storage['smbConfig'] = {
+            'server': storage['server'],
+            'share': storage['share'],
+            'domain': storage['domain'],
+            'username': storage['username'],
+            'password': storage['password'],
+            'size': storage['size'],
+        }
+        delKeys.extend(['server', 'share', 'domain', 'username', 'password', 'size'])
```
Qubad786, line 44: python uses snake case i.e `delete_keys`

library/ix-dev/community/n8n/migrations/migrate
```diff
@@ -0,0 +1,85 @@
...
+        check_val = values.get(storageKey, {}).get(storage, {})
+        if not isinstance(check_val, dict) or not check_val:
+            continue
+
+        values[storageKey][storage] = storageMigrate(check_val)
+
+
+    additionalStorages = values.get(storageKey, {}).get('additionalStorages', [])
+    for idx, storage in enumerate(additionalStorages):
+        if not isinstance(storage, dict) or not storage:
```
Qubad786, line 67: So if for example `additionalStorages` had 3 items - for index 0 the condition was false and for index 1 it was true and for index 2 is false again.
Why do we keep the inconsistent entry still at index 1?

stavros-k: I think we can safely remove the if check here (and above in L59, data, pgData and pgBackup cannot be empty or not a `dict`).
Both previous and current schema requires values to be set (or the item to _not_ exist), so I dont expect malformed data there.

Qubad786: @stavros-k what i am trying to say is that if we have malformed data anywhere which is not expected..we should likely raise an exception and not do anything at all because it can't be malformed if it was not manipulated directly

sonicaj (approved):

PR #1772, opened 2023-11-21
immich - adapt to upstream changes

stavros-k:


library/ix-dev/community/immich/Chart.yaml
```diff
@@ -3,9 +3,9 @@ description: Immich
 annotations:
   title: Immich
 type: application
-version: 1.1.0
+version: 2.0.0
```
stavros-k, line 6: This is not _really_ a breaking change.
But chart-test will use _current_ templates with _previews_ values.yaml (which includes older images) as part of upgrade test.
That will never work, cause of the changes. Bumping major versions makes CI skip this step.

TL;DR is breaking for the CI only.
Users will not have any breaking changes

Chaphasilor: > Users will not have any breaking changes

We will see 😁  
But I have faith in your work! <3

Qubad786 (approved):

PR #1805, opened 2023-11-28
Enterprise/Syncthing - Additional envs

stavros-k:


library/ix-dev/enterprise/syncthing/questions.yaml
```diff
@@ -31,6 +31,74 @@ questions:
...
         - definitions/timezone
 
+  - variable: syncthingConfig
+    label: ""
+    group: Syncthing Configuration
+    schema:
+      type: dict
+      attrs:
+        # Boolean flags are currently
+        # not exposed to the user
```
stavros-k, line 41: Flags added as **hidden** here because are already exist in the values.yaml to avoid issues with schema.

sonicaj (approved):

PR #1858, opened 2023-12-11
remove immutable from hostPath

### LINKED ISSUE #1872: Additional storage greyed out for deluge and jellyfin
When I try to add additional storage for deluge and jellyfin the "type" dropdown is greyed out and unclickable. I can type in the mount path though.

┆Attachments: <a href="https://api.atlassian.com/ex/jira/94e022be-3595-4f54-979f-780bfeff904d/rest/api/2/attachment/content/60874">debug-truenas-20231209104703.tgz</a>

stavros-k:
https://ixsystems.atlassian.net/browse/TNCHARTS-363

CI Fails are expected.
Its because the way it tests the upgrade, (it tries to use the old values with new templates).
Which in our case does not happen on real systems.

Closes #1872


library/ix-dev/community/searxng/questions.yaml
```diff
@@ -231,9 +230,7 @@ questions:
                             schema:
                               type: hostpath
                               show_if: [["aclEnable", "=", false]]
-                              immutable: true
-                              required: true
-                    - variable: ixVolumeConfig
+                              required: trueig
```
stavros-k, line 233:
```suggestion
                              required: true
                    - variable: ixVolumeConfig
```

Qubad786 (commented):
I was just curious why we need to explicitly set `emptyDirConfig` to use default values..

stavros-k:
> I was just curious why we need to explicitly set `emptyDirConfig` to use default values..

I've explicitly made it required in common as a sanity check during all those migrations. 
Once we clear things up we can remove that indeed

Issue #1848, opened 2023-12-07
Immich Upcoming Breaking Change


Chaphasilor:
Just FYI, Immich v1.91 will have another breaking change that removes one container completely and changes the base image of another one.

The change is described in the v1.90 release notes here:  
https://github.com/immich-app/immich/releases/tag/v1.90.0

jbgosselin:
The new release just dropped with the breaking change https://github.com/immich-app/immich/releases/tag/v1.91.0

PR #1907, opened 2023-12-18
Immich upstream breaking changes

stavros-k:
Closes https://github.com/truenas/charts/issues/1848

Adapts to upstream breaking changes https://github.com/immich-app/immich/releases/tag/v1.91.0

Chances of an actual breaking change is almost zero, but users should proceed with caution (Hence the major chart version bump)


library/ix-dev/community/immich/templates/_postgres.tpl
```diff
@@ -1,7 +1,23 @@
...
+    {{- $_ := set .Values "pgvectorImage" (dict
+      "repository" "tensorchord/pgvecto-rs"
+      "pullPolicy" "IfNotPresent"
+      "tag" "pg15-v0.1.11"
+    ) -}}
+  {{- end }}
+
+{{- $psql := "PGPASSWORD=${POSTGRES_PASSWORD} psql --host=${POSTGRES_HOST} --dbname=${POSTGRES_DB} --username=${POSTGRES_USER}" -}}
+{{- $tasks := (list
+  (printf "%s -c \"ALTER DATABASE ${POSTGRES_DB} REFRESH COLLATION VERSION;\"" $psql)
```
stavros-k, line 14: this is needed due the slight downgrade from pg 15.2 to pg 15.0 (the only image/version that pgvector has available on the same major version).

Thankfully its not very expensive (on my testing it was instant) and its does not fail or re-run if there isn't something to refresh

Qubad786 (approved):

PR #1911, opened 2023-12-19
bump common

stavros-k:


Qubad786 (approved):
I presume that test fails for Zerotier have been taken into consideration.

stavros-k:
> I presume that test fails for Zerotier have been taken into consideration.

yes, we don't have an actual api key, we just observe logs for anomalies

PR #1931, opened 2023-12-20
fix wg-easy interface

### LINKED ISSUE #1369: WG-Easy App: Fix Hardcoded Device Name

### LINKED ISSUE #1877: wg-easy app assumes network interface name is eth0
the wg-easy app in the iX catalog uses the default network interface name of eth0. Most Truenas systems have different name. This name is used to set up the route tables. Specifically  the Masquerade route entry to the network interface.

Effectively, this prevents access to the remaining lan (and internet).

The only way to fix this is to add a environemental command entry of 

WG_POST_UP = 

iptables -t nat ~~A POSTROUTING~~ s 10.8.0.0/24 ~~o br0~~ j MASQUERADE; iptables ~~A INPUT ~~p udp ~~m udp~~ --~~-dport 51820~~ j ACCEPT; iptables ~~A FORWARD~~ i wg0 ~~j ACCEPT; iptables~~ A FORWARD ~~o wg0~~ j ACCEPT; 

Where “br0” is the network interface name.

Suggest you add a field to define the Network interface name or use the WG_DEV environmental variable.

Also because iptable commands keep amending the list gets very long. Consider clearing the iptables entry at startup.

### LINKED ISSUE #1930: wg-easy moved container repo
wg-easy moved container repo docker Hub weejewel/wg-easy has been moved to ghcr.io/wg-easy/wg-easy

it would seem that v8 and future updates will not available on docker hub

source: https://hub.docker.com/r/weejewel/wg-easy#!

stavros-k:
Adds the option that was added here https://github.com/wg-easy/wg-easy/pull/500

Fixes #1877 
Fixes #1369 
Fixes #1930 
Covers #1370

### DIFF
--- a/library/ix-dev/charts/wg-easy/Chart.yaml
+++ b/library/ix-dev/charts/wg-easy/Chart.yaml
@@ -3,9 +3,9 @@ description: WG-Easy is the easiest way to install & manage WireGuard!
 annotations:
   title: WG Easy
 type: application
-version: 2.0.4
+version: 2.0.5
 apiVersion: v2
-appVersion: "7"
+appVersion: "8"
 kubeVersion: ">=1.16.0-0"
 maintainers:
   - name: truenas
--- a/library/ix-dev/charts/wg-easy/questions.yaml
+++ b/library/ix-dev/charts/wg-easy/questions.yaml
@@ -111,6 +111,13 @@ questions:
             type: int
             required: true
             default: 0
+        - variable: deviceName
+          label: Device Name
+          description: Ethernet device the wireguard traffic should be forwarded through.
+          schema:
+            type: string
+            required: true
+            default: eth0
         - variable: clientMTU
           label: Clients MTU
           description: The MTU the clients will use.
--- a/library/ix-dev/charts/wg-easy/templates/_wgeasy.tpl
+++ b/library/ix-dev/charts/wg-easy/templates/_wgeasy.tpl
@@ -24,6 +24,7 @@ workload:
                 - SYS_MODULE
           env:
             WG_PORT: {{ .Values.wgNetwork.udpPort }}
+            WG_PATH: /etc/wireguard
             PORT: {{ .Values.wgNetwork.webPort }}
             WG_HOST: {{ .Values.wgConfig.host | quote }}
             PASSWORD: {{ .Values.wgConfig.password | quote }}
@@ -31,6 +32,7 @@ workload:
             WG_MTU: {{ .Values.wgConfig.clientMTU }}
             WG_DEFAULT_ADDRESS: {{ .Values.wgConfig.clientAddressRange }}
             WG_DEFAULT_DNS: {{ .Values.wgConfig.clientDNSServer }}
+            WG_DEVICE: {{ .Values.wgConfig.deviceName | default "wg0" }}
             WG_ALLOWED_IPS: {{ join "," .Values.wgConfig.allowedIPs | default "0.0.0.0/0,::/0" | quote }}
           fixedEnv:
             PUID: 0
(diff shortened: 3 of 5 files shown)


JamesOsborn-SE:
Is there a reason that `WG_MTU` is required when the docs show it as having a default of `null`?

stavros-k:
> Is there a reason that `WG_MTU` is required when the docs show it as having a default of `null`?

Correct me if I'm wrong, but wireguard itself will default to 1420 if you don't provide a value.
By requiring it, user knows exactly value is always used.

Is there a use case that null would change something that I'm missing?

JamesOsborn-SE:
> > Is there a reason that `WG_MTU` is required when the docs show it as having a default of `null`?
> 
> Correct me if I'm wrong, but wireguard itself will default to 1420 if you don't provide a value. By requiring it, user knows exactly value is always used.
> 
> Is there a use case that null would change something that I'm missing?

`null` would change the user experience.
A default that isn't required means it should work without the user providing input; I think that's true here. Since the service doesn't require it why should the chart?
If it's not required maybe it isn't passed as `null` to the service.

But I don't know what this line does so I thought it might be required:
https://github.com/wg-easy/wg-easy/blob/83ac4ff4cf96b23ec4e1d8019413311834083d7a/src/lib/WireGuard.js#L204

I'm try to understand why it's this way; there may be no action needed.

stavros-k:
> > > Is there a reason that `WG_MTU` is required when the docs show it as having a default of `null`?
> > 
> > 
> > Correct me if I'm wrong, but wireguard itself will default to 1420 if you don't provide a value. By requiring it, user knows exactly value is always used.
> > Is there a use case that null would change something that I'm missing?
> 
> `null` would change the user experience. A default that isn't required means it should work without the user providing input; I think that's true here. Since the service doesn't require it why should the chart? If it's not required maybe it isn't passed as `null` to the service.
> 
> But I don't know what this line does so I thought it might be required: https://github.com/wg-easy/wg-easy/blob/83ac4ff4cf96b23ec4e1d8019413311834083d7a/src/lib/WireGuard.js#L204
> 
> I'm try to understand why it's this way; there may be no action needed.

User don't have to provide something either now, the default is `1420` (the same that wireguard, will use if not provided)

> Note, I say `wireguard` not `wg-easy` in the above statement

empty field even as optional will have the user looking at it and thinking if this should be set to something or not.

JamesOsborn-SE:
I think this will still be seen as "broken" by users that expect to connect to the internet via clients connecting to wg-easy. I think this is an upstream issue I made a comment here: https://github.com/wg-easy/wg-easy/issues/562#issuecomment-1866712931

You may want to provide a warning in the documentation or drop support to community to manage expectations.

Qubad786 (approved):

stavros-k:
> I think this will still be seen as "broken" by users that expect to connect to the internet via clients connecting to wg-easy. I think this is an upstream issue I made a comment here: [wg-easy/wg-easy#562 (comment)](https://github.com/wg-easy/wg-easy/issues/562#issuecomment-1866712931)
> 
> You may want to provide a warning in the documentation or drop support to community to manage expectations.

I've gone ahead and merged this is as is for now, as it fixes the interface issue, as what you mention here, feel free to create a ticket and we can look into that.
Thanks

PR #2071, opened 2024-01-23
Allow Volumes to be mounted in Tailscale to avoid machine name duplication on chart update

CharlesSaint007:
Add option to allow volumes to be mounted to allow the following volumes to be mounted:

      - '/dev/net/tun:/dev/net/tun'
      - '/var/lib:/var/lib'

Mounting these volumes allows tailscale to be updated WITHOUT causing the machine name to be duplicated. As it stands today, EVERY TIME the chart is updated, it causes the machine to be duplicated (and the name changed since the machine already exists). eg. If the machine name is 'server' and the chart is updated, when the chart re-deploys, it tries to register a new machine as 'server', fails to do so (since that machine already exists), and instead deploys as 'server-1'. This is a major annoyance.

### DIFF
--- a/community/tailscale/1.0.31/ix_values.yaml
+++ b/community/tailscale/1.0.31/ix_values.yaml
@@ -21,3 +21,9 @@ tailscaleConfig:
 
 tailscaleNetwork:
   hostNetwork: false
+
+config:
+  type: ixVolume
+  ixVolumeConfig:
+  datasetName: config
+  additionalStorages: []


stavros-k (changes requested):
Hello,

1. Please only alter files under `library/ix-dev` directories
2. The definition in the values file its not even a valid yaml syntax
3. The definition is not plumbed in the templates
4. The options are not exposed to the UI (questions.yaml).

But beyond that, tailscale in kubernetes uses Kubernetes Secrets to keep state.
Even if the a volume is added, it won't do anything, at least based on the upstream docs. Can you provide more info/sources why would this help?

CharlesSaint007:
> Hello,
> 
> 1. Please only alter files under `library/ix-dev` directories
> 2. The definition in the values file its not a valid yaml syntax
> 3. The definition is not plumbed in the templates
> 4. The options are not exposed to the UI (questions.yaml).
> 
> But beyond that, tailscale in kubernetes uses Kubernetes Secrets to keep state. Even if the a volume is added, it won't do anything, at least based on the upstream docs. Can you provide more info/sources why would this help?

Sorry for the syntax issues and putting the changes in the wrong place @stavros-k . I have Tailscale running via Docker on other Linux machines, and adding those volumes solved my issue with duplicate machines after every update. If Kubernetes Secrets is expected to keep the state, then I guess my question is why doesn't it work? Right now, EVERY TIME you update the app, it duplicates the machine name instead of using the existing machine. So I guess, if adding the aforementioned volumes isn't the fix... Then what is?

CharlesSaint007:
I think fixed the issue using an environment variable instead

PR #2119, opened 2024-02-01
netboot.xyz - Match values.yaml to questions.yaml

### LINKED ISSUE #2114: netbootxyz - Can't update because of host network ports
I attempted to update netboot today and got the error below because of the host network ports.  I was able to remove and install new version with the same configuration, but it would not do in-place upgrade.  Is this a bug in TrueNAS itself, maybe something with update is checking things it shouldn't???, or something we missed in the chart?

Error: [EINVAL] values.netbootNetwork.tftpPort: Should be between 9000 and 65535
[EINVAL] values.netbootNetwork.webHttpPort: Should be between 9000 and 65535

CompPhy:
Fixes #2114


library/ix-dev/community/netbootxyz/Chart.yaml
```diff
@@ -3,7 +3,7 @@ description: netboot.xyz lets you PXE boot various operating system installers o
 annotations:
   title: netbootxyz
 type: application
-version: 1.0.1
+version: 1.0.2
 apiVersion: v2
 appVersion: 2.0.76
```
stavros-k, line 8:
```suggestion
appVersion: 0.7.0.4
```

library/ix-dev/community/netbootxyz/values.yaml
```diff
@@ -1,7 +1,7 @@
 image:
   repository: netbootxyz/netbootxyz
   pullPolicy: IfNotPresent
-  tag: 0.7.0-nbxyz2
+  tag: 0.7.0-nbxyz4
```
stavros-k, line 4: Can you also make upgrade_strategy executable?

`chmod +x upgrade_strategy`

stavros-k (approved):

Qubad786 (approved):

PR #2129, opened 2024-02-02
netboot.xyz - WEB_APP_PORT to configure the web application to the correct port.

### LINKED ISSUE #2128: netboot.xyz - Portal link does not work correctly when host network is enabled.
I'm going to enter a PR for this after entering this description.  I have proposed an upstream change which makes the web app port fully configurable within the container image.  This PR will remain as WIP pending approval of the upstream change.  netbootxyz/docker-netbootxyz#57

We setup the chart to use variables for TrueNAS port information.  Based on the current setup, if host network is enabled the portal gets bound on port 3000, not configurable.  However, the URL linked in TrueNAS uses the values.yaml which is defaulting to a higher port number.  This results in a broken link.

The current work around is to just put in port 3000 manually when accessing the web app in this configuration.  Ideally the container should just use the right port here.

CompPhy:
Fixes  #2128

Will remain as draft pending upstream approval of netbootxyz/docker-netbootxyz#57


CompPhy:
I do realize we should bump the version number here.... I will do that once the change is possible to merge.

stavros-k (approved):

stavros-k (approved):

stavros-k (approved):

Qubad786 (approved):

Issue #2078, opened 2024-01-25
Add GPU configuration to nextcloud


frederickjansen:
Certain Nextcloud plugins can use the GPU for hardware acceleration. [Memories](https://github.com/pulsejet/memories) for example uses it to transcode video, and [Recognize](https://apps.nextcloud.com/apps/recognize) for its machine learning models. GPU selection should be added as an option to the Nextcloud container.

PR #2161, opened 2024-02-12
Add Nextcloud GPU config

frederickjansen:
Not sure if this is the correct format to add GPU support to a chart, but if it is this would close https://github.com/truenas/charts/issues/2078.


stavros-k:
This is not enough sadly, Nextcloud is still on the old library, which would need more work to add GPU that easily.
I'm in the process of migrating last few apps to the new library, So I'd suggest to sit tight! Shouldnt take long

frederickjansen:
I figured it wasn't that easy! Thanks for working on this, much appreciated.

PR #2172, opened 2024-02-14
re-adds photoprism 1.1.3

### LINKED ISSUE #2171: photoprism upgrade not possible
Last photoprism version was 231128_1.1.2 but 231128_2.2.0 requires 1.1.3 to allow upgrade

" [EFAULT] Failed to upgrade App: Migration to 2.x.x is only allowed from 1.1.3 or higher "

stavros-k:
Fixes #2171

Restore the photoprism v1.1.3 so users can upgrade to that first and then to the v2.x.x


Qubad786 (approved):

library/ix-dev/charts/photoprism/to_keep_versions.md
```diff
@@ -0,0 +1,4 @@
+# 1.1.13
+
+This version is kept because it contains a fix that is needed for migration to v2.x.x
+It should be safe to remove few months after v2.x.x is released.
```
breadtk, line 4: Might want to give a specific date/month to remove the version. IMHO something conservative like 6months from today should be more than sufficient.

PR #2178, opened 2024-02-16
n8n bump version

stavros-k:
n8n lately publishes pre-releases with the same versioning format as the stable ones, 
Lets bump for now manually to the latest stable release and hope the situation gets better

### DIFF
--- a/library/ix-dev/community/n8n/Chart.yaml
+++ b/library/ix-dev/community/n8n/Chart.yaml
@@ -3,9 +3,9 @@ description: n8n is an extendable workflow automation tool.
 annotations:
   title: n8n
 type: application
-version: 1.2.21
+version: 1.2.22
 apiVersion: v2
-appVersion: 1.25.1
+appVersion: 1.25.3
 kubeVersion: '>=1.16.0-0'
 maintainers:
   - name: truenas
--- a/library/ix-dev/community/n8n/Chart.yaml
+++ b/library/ix-dev/community/n8n/Chart.yaml
@@ -5,7 +5,7 @@ annotations:
 type: application
 version: 1.2.22
 apiVersion: v2
-appVersion: 1.25.3
+appVersion: 1.27.3
 kubeVersion: '>=1.16.0-0'
 maintainers:
   - name: truenas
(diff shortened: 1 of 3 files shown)


Qubad786 (approved):

netroy:
> n8n lately publishes pre-releases with the same versioning format as the stable ones

This is actually not true. If a n8n docker tag follows semver, then the release is a stable one.

It's unfortunate that you folks were affected by https://github.com/n8n-io/n8n/issues/8664, but running the application with a custom UID/GID isn't something we've **ever** officially supported, and **there are no checks in place to prevent that case from breaking again**.

Obviously I can't tell you what UID/GID to run your pods as, but if someone starts running n8n on deno/bun, and when every now and then the application breaks because we only support node.js, is that really the fault of the n8n maintainers?

That said, the "fix" for that issue was merged 2 days ago, and will be released in `1.30.0` sometime later today. Once we upgrade that version (or a later version), you can remove the `SHELL` env variable from the chart. 
However, I'd still highly recommend (for any application) checking with project maintainers if their application supports custom UID/GID, before doing so in a chart.

stavros-k:
Hello, let me make clear that the UID/GID issue has nothing to do with the tags. 
It just happened to add the fix here while bumping the tag.

> This is actually not true. If a n8n docker tag follows semver, then the release is a stable one.

Thanks for making clear how the tagging works, because it was confusing looking at github releases and see-ing a pre-release but docker tag looking like a stable one! If this was something in the docs explaining this I've probably missed it.

> It's unfortunate that you folks were affected by [n8n-io/n8n#8664](https://github.com/n8n-io/n8n/issues/8664), but running the application with a custom UID/GID isn't something we've **ever** officially supported, and **there are no checks in place to prevent that case from breaking again**.

Don't worry about it, there will be always bugs, this is why CI is in place to catch _unexpected_ issues like the UID/GID you have mentioned. 

Which I utilized to bump the tag and try a bunch of things to see where the issue originates from.
Then I've reported and you provided a solution + fix.

Thanks

netroy:
> it was confusing looking at github releases and seeing a pre-release but docker tag looking like a stable one! If this was something in the docs explaining this I've probably missed it.

Yeah. Sorry about that bit. We decided at some point to start using the `pre-release` flag to deter people's setups from auto-upgrading seconds after the release was made. We are definitely not using that flag correctly, but we still need to keep using it for now because in the current release process the GitHub release is created about 5 minutes before the docker image is available.
I've created an internal ticket to move away from this practice, and hopefully in the near future we can completely stop using the `pre-release` flag.

PR #2212, opened 2024-02-22
n8n - change execution mode

### LINKED ISSUE #2210: n8n workers not started
It seems currently the `n8n` chart [uses "queue mode"](https://docs.n8n.io/hosting/scaling/queue-mode/) but only starts the main process, not the workers. 

``` bash
/home/node $ ps | grep n8n
    7 568       6:35 node /usr/local/bin/n8n
   48 568       0:00 grep n8n
```

As a result the “test workflow” button works, but any production executions will get stuck because there are no workers to pick them up.

A possible solution is to configure `n8n` to run in "regular mode" (i.e., just 1 process), or we need to start at least 1 worker.

See also: https://community.n8n.io/t/workflow-stuck-in-running-state/40473/4

stavros-k:
Fixes #2210 

- Changes `EXECUTION_MODE` to `regular`, as the current `queue` value is only useful if you have additional workers to pick up the work. 
- Removes a temporary fix that is no longer needed on the newer container tag.

### DIFF
--- a/library/ix-dev/community/n8n/Chart.yaml
+++ b/library/ix-dev/community/n8n/Chart.yaml
@@ -3,7 +3,7 @@ description: n8n is an extendable workflow automation tool.
 annotations:
   title: n8n
 type: application
-version: 1.2.24
+version: 1.2.25
 apiVersion: v2
 appVersion: 1.30.0
 kubeVersion: '>=1.16.0-0'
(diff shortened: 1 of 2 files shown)


yichi-yang:
I think Redis is only used in the `queue` mode? Should we get rid of that pod now that we changed `EXECUTION_MODE` to `regular`?

Issue #2134, opened 2024-02-03
Community Chart Suggestion:MoviePilot


gaodaxiaya:
https://github.com/jxxghp/MoviePilot

PR #2216, opened 2024-02-23
add moviepilot

gaodaxiaya:
add moviepilot #2134


stavros-k:
Hello, not sure if you have completed your work or you still working on it, just want to avoid, you going down the wrong path!

- Looks like you haven't added any templates yet, which is the only think that makes it to actually work
- Please only edit/add files under `library/ix-dev/` directory. In this case `/library/ix-dev/community/moviepilot`
- Follow the format of other apps in the community train, as I've noticed based on your questions.yaml that you probably used one of the apps that are still on the old library
- Chart.yaml is missing the `dependencies` which you will need to have the `common` library referenced (so it can actually work)

Please ignore any of the points above if you already plan to work on them, just didn't want you to spent time on a wrong direction!

Let me know if you have any questions.

Thanks!

PR #2250, opened 2024-03-04
dashy - fix root fs ro

stavros-k:


stavros-k:
not needed anymore, seems to be fixed upstream

PR #2267, opened 2024-03-08
set secure cookie based on the cert

### LINKED ISSUE #2262: n8n app changed to pre-release, breaks usage
Update to the n8n chart switched to the pre-release 1.32 version, which is not stable but pre-release. Updating the app breaks login to the app via the web portal.

This should be downgraded to the latest stable 1.31.x release: 1.31.2
https://github.com/n8n-io/n8n/releases

stavros-k:
Fixes #2262


Joffcom:
Looks good to me.

Qubad786 (approved):

PR #2377, opened 2024-04-12
nextcloud migration

stavros-k:


frederickjansen:
@stavros-k could you please add this as well if you haven't already? https://github.com/truenas/charts/issues/2078
Thanks!

stavros-k:
> @stavros-k could you please add this as well if you haven't already? #2078 Thanks!

I can do right after this PR, but I believe that nextcloud container does not have the required packages installed by default for that to work.

Qubad786 (approved):

PR #2479, opened 2024-05-15
webdav: allow modifying the request body limit

### LINKED ISSUE #2462: WebDAV Write Access Issue
I created a WebDAV service and mounted it using RaiDrive on Windows 10.
 I can read and copy small files normally, but when copying large files, I get a permission denied error. I suspect it may be because the size of a single file exceeds 3.6 GB. So, I want to enter the container to modify /usr/local/apache2/conf/Includes/webdav-ssl.conf, but I'm getting a permission denied error and can't obtain the temporary root password.

I tried uploading a single file using FileBrowser, and there were no issues. Additionally, everything works fine via SFTP as well.

### LINKED ISSUE #2478: WebDAV app permit to upload files up to 200-250MB. Over that size I receive a error
Hi,

I create a webdav app (with HTTPS enabled; HTTP disabled) using a dataset like share mount point anche a dedicated user by Basic Authentication.

I configured setting in the GUI and using a webdav client (like winSCP) it seems work fine.

I can access the share, create dir, read files and even upload small files.

The issue is that when I try to upload files greater than 200-250MB the transfer goes in error. There is no way to upload that file.

Looking at pod log I can see:

2024-05-15 12:42:13.692422+02:00[Wed May 15 10:42:13.691885 2024] [dav:error] [pid 9:tid 140661650601664] (-102)Unknown error ~~102: [client x.x.x.x:58407] An error occurred while reading the request body (URI: /SSU/2022~~04-17-ubuntu-16.04-mate-desktop-mpv-1080p-bpi-m2z-sd-emmc.img.zip) [-102, #0]

I’m using TrueNAS-SCALE-23.10.2

Webdav app info:

Name:  webdav

App Vesion:  2.4.59

Chart Version:  1.0.23

stavros-k:
According to apache does the default LimitRequestBody is 1GB
https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestbody

I've left the default as `1`, but now user can specify per share. 

Fixes #2478 
Fixes #2462

### DIFF
--- a/library/ix-dev/community/webdav/Chart.yaml
+++ b/library/ix-dev/community/webdav/Chart.yaml
@@ -4,7 +4,7 @@ description: WebDAV is a set of extensions to the HTTP protocol which allows use
 annotations:
   title: WebDAV
 type: application
-version: 1.0.23
+version: 1.0.24
 apiVersion: v2
 appVersion: 2.4.59
 kubeVersion: '>=1.16.0-0'
--- a/library/ix-dev/community/webdav/ci/http-basicauth-values.yaml
+++ b/library/ix-dev/community/webdav/ci/http-basicauth-values.yaml
@@ -17,6 +17,7 @@ webdavStorage:
       description: My third disabled share
       hostPath: /mnt/{{.Release.Name }}/share3
       readOnly: true
+      maxRequestBodySizeInGB: 10
       fixPermissions: false
 
 webdavNetwork:
--- a/library/ix-dev/community/webdav/questions.yaml
+++ b/library/ix-dev/community/webdav/questions.yaml
@@ -180,8 +180,8 @@ questions:
                         Example: [share1] will be available at [http://<webdav-ip>:<webdav-port>/share1]
                       schema:
                         type: string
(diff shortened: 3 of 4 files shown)


Qubad786 (approved):

ros3dprinter:
Hi,

I tested new version but the problem still exists.
I set the max file size to 0 (unlimited) but I still receive Entity too large error
immagine
Senza titolo

stavros-k:
Do you use any kind of proxy infront of the webdav?

ros3dprinter:
No. I'm connecting directly to webdav

ros3dprinter:
I did a new test thinking the problem may be the 0 value.

I set the parameter to 2 (that means 2GB) and the upload worked fine.

So the problem seems to be that it doesn't set properly only the 0 value /which means unlimited).


immagine

immagine

ros3dprinter:
I tried to upload a file greater than 2GB and I obviously get error.

So the fix just need to manage properly the unlimited value.

immagine

stavros-k:
Whoops, you are right.
Fixed here:
https://github.com/truenas/charts/pull/2484

Thanks!

PR #2480, opened 2024-05-15
Update audiobookshelf version

joeySchentrup:


joeySchentrup:
Thank you for the quick response!

PR #2484, opened 2024-05-16
webdav: handle 0 in limitReqBody correctly

stavros-k:


ros3dprinter:
Hi,
can you specify when new code will be available?  Version will reach 1.0.25 ?

Qubad786 (approved):

Qubad786:
> Hi, can you specify when new code will be available? Version will reach 1.0.25 ?

Yes, it will be available as soon as this PR is merged i.e version 1.0.25

ros3dprinter:
Hi,

I tested new 1.0.25 and now it works fine. I was able to set 0 (unlimited) and transfer file greater than 1GB (last test was with 3.2GB).

thanks for your support

Regards

PR #2575, opened 2024-06-14
add config option for jellyfin timezone

### LINKED ISSUE #2574: jellyfin container cannot configure timezone
the jellyfin chart does not have a question for setting the timezone. This means the common tz UTC gets set and cannot be overridden by e.g. an environment variable or mounting /etc/localtime.

I  am going to submit an untested PR that merely copies a timezone questions block that I saw in many other community charts in this repo. Could perhaps someone with a dev environment set up test whether it works? :fade:

deargle:
Attempts to maybe possibly fix #2574. untested! copies timezone question blocks I saw in other community charts.

### DIFF
--- a/community/jellyfin/1.2.16/questions.yaml
+++ b/community/jellyfin/1.2.16/questions.yaml
@@ -21,6 +21,16 @@ portals:
     path: "$kubernetes-resource_configmap_portal_path"
 
 questions:
+  - variable: TZ
+    group: Jellyfin Configuration
+    label: Timezone
+    schema:
+      type: string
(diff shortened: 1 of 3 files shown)


deargle:
Done, thanks for guidance and testing! I bumped to 1.3.0 following semver since this is a feature-add, but I can also see reasoning for it being a bugfix, lmk if you prefer a bump 1.2.17 instead.

stavros-k (approved):
Looks good, thanks!

Qubad786 (approved):

PR #2583, opened 2024-06-17
tdarr: fix upgrade regex

swensorm:
Noticed this morning that Tdarr has not upgraded for several versions.  I believe it's due to the regex having `_ffmpeg5` as a required suffix.  Tdarr has suffixed `_ffmpeg5` on some image tags but not the most recent, and sometimes used 4 version segments according to the tag history on DockerHub; https://hub.docker.com/r/haveagitgat/tdarr/tags.

Updated the regex in the upgrade_strategy to support both scenarios.  Also removed the string quotes on the tag - likely not the issue but consistent with other applications.

Please let me know if you would like an Issue ticket for this as well.  Thank you very much!

### DIFF
--- a/library/ix-dev/community/tdarr/Chart.yaml
+++ b/library/ix-dev/community/tdarr/Chart.yaml
@@ -3,9 +3,9 @@ description: Tdarr is a Distributed Transcoding System
 annotations:
   title: Tdarr
 type: application
-version: 1.2.3
+version: 1.2.4
 apiVersion: v2
-appVersion: '2.17.01'
+appVersion: 2.17.01
 kubeVersion: '>=1.16.0-0'
 maintainers:
   - name: truenas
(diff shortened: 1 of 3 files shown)


stavros-k (approved):
I'm not 100% sure that semantic versioning func takes into account the 4th number/segment, but overall looks good!

Thanks!

swensorm:
Thanks @stavros-k!  TBH I was confused by python's docs as some specify only 3 (https://packaging.pypa.io/en/stable/version.html) but others show the release segment with an arbitrary wildcard of digits (https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers), so I tested out in a python:3 container and looks to be parsing all fo them into a tuple:
```py
>>> parse_version('1.2.3.4.5.6')
<Version('1.2.3.4.5.6')>
>>> parse_version('1.2.3.4.5.6') > parse_version('1.2.3.4.5.5')
True
```

Found other apps like Collabora that have 5 in their regex, and emby with 4 so I _think_ it should work 🙏

Qubad786 (approved):

PR #2678, opened 2024-07-23
add `playwright` to `community` train

swensorm:
Hello,

Was not sure if you are still allowing new applications to be added with the docker compose news I just read about, but hoping to have [Playwright](https://github.com/truenas/charts/issues/2680) considered for inclusion in the catalog.  A bit unfortunate as I was hoping User Namespaces were going to be included soon which would fix a lot of compatibility issues.  Playwright has a UI mode that runs in a container and inclues Chrome, Firefox, and Webkit inside.  Useful for both UI testing and automated browser tasks scheduled with Cron 😄 

Playwright does not have an official volume mapping but it doesn't do anything to run the container without a test project as there will be nothing to run, so I've forced a basic `/project` persistence folder.  It will only make sense to use hostPath or SMB but left it standard for consistency with others.  It is also needed to set the working directory to run the npx command to start the ui.  I have this today with the ix-chart doing '/bin/bash -c "cd /project && npx -y playwright test ..". but trying with `workingDir` in the chart since it's available.  Open to changing this though.

If open to inclusion, I would greatly appreciate if the svg could be copied to iX systems media server (`https://media.sys.truenas.net/apps/playwright/icons/icon.svg`) from https://github.com/microsoft/playwright/blob/cf1a313a0c5f39caf2a6be360ce5790749a87c9f/packages/recorder/public/playwright-logo.svg?plain=1

And if the following png images could be included for screenshots 1-6

- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266656846-6b87712f-64a5-4d73-a91d-6562b864712c.png
- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266684061-811a9985-32aa-4a3e-9869-de32053cf468.png
- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266691440-f9f43a0c-78d7-4574-9a58-c69d2ec53c8f.png
- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266684395-9e7eeb84-bd26-4010-8614-75e24b56c716.png
- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266691884-49b9fa2a-8a57-4044-acaa-0a2ea4784c5c.png
- https://github-production-user-asset-6210df.s3.amazonaws.com/13063165/266688701-ffca2fd1-5349-41fb-ade9-ace143bb2c58.png

These were taken from the UI Mode page; https://playwright.dev/docs/test-ui-mode

There is a bit to this that's atypical compared to other apps in the catalog so I'll understand any eyebrow raises 😂 

Thank you very much for the consideration and review!


swensorm:
It appears that unfortunately the `/trace/ping` route is only permitted by the Service Worker.  I'm not sure if this would support keeping a SW established and connected for a persistent ping.  We could request the logo file at `/trace/playwright-logo.svg` but not sure that's better.  I've changed it to the root in the meantime; `/`.

library/ix-dev/community/playwright/questions.yaml
```diff
@@ -0,0 +1,367 @@
...
+    schema:
+      type: dict
+      attrs:
+        - variable: user
+          label: User ID
+          description: The user id that Playwright will run as.
+          schema:
+            type: int
+            min: 2
+            default: 568
```
swensorm, line 65: Playwright adds a non-root user `pwuser` which is 1000 to the image (https://github.com/microsoft/playwright/blob/main/utils/docker/Dockerfile.jammy#L26).  If running standalone on an empty folder (like for the CI build) then it has to run as 1000, but it runs fine to be configured as the standard `apps` user when running on an actual playwright test project.  So, in my last push I defaulted to 1000 for running but if the user does choose to configure the `playwrightRunAs` it will be 568 by default like other apps.

Please let me know if you would prefer this to be 1000 too though.

stavros-k:
Hello, I'd be more in favor of waiting and adding this on the new system that will use docker in `truenas/apps` repo.
Currently there are not something that you can use as a template, but once the release is out, you can take a look of released apps.

Idea is that if this is merged now, then we would have to also add it on the truenas/apps repo AND add custom migration for it.

swensorm:
Ok, no worries.  With how many apps there are and frequent updates/fixes was thinking there may have been an automated conversion.  I have this running ok in ix-chart currently so not an  urgency.  I'll close this and wait for next year.

PR #2686, opened 2024-07-25
Add threadfin to community train

Soodrax:
Hello,

This request is to add threadfin to the community train as mentioned here [2656](https://github.com/truenas/apps/issues/271.). I've already attempted to create a PR but it was rightly pointed out that I was doing it in the wrong way or in fact where i was trying to contribute to was incorrect. Thanks @swensorm for pointing me in the right direction. Please find my revised PR and hopefully I've done this correctly this time.

Thanks.


stavros-k:
Hello, I'd be more in favor of waiting and adding this on the new system that will use docker in `truenas/apps` repo.
Currently there are not something that you can use as a template, but once the release is out, you can take a look of released apps.

Idea is that if this is merged now, then we would have to also add it on the truenas/apps repo AND add custom migration for it.

Soodrax:
Sounds good to me.

stavros-k:
Hey, I'm going to close this now, I've already moved the app request here https://github.com/truenas/apps/issues/271 if you want to subscribe, (if not already).

Thanks!

PR #2767, opened 2024-08-26
nextcloud: fix portal

stavros-k:


Qubad786 (commented):
Why are we updating `catalog.json`?

stavros-k:
> Why are we updating `catalog.json`?

Needed a rebase!
Fixed

PR #2780, opened 2024-08-30
frigate: Increase max cache size 

mtthidoteu:
Facing cache issues, therefore it would be useful to be able to increase this past 4GB


stavros-k (changes requested):
Hello, please only apply changes under `ix-dev` directory.

The versioned directories are auto genearted.

Thanks

stavros-k:
Should be fixed for frigate app in truenas/apps repo.

PR #2862, opened 2024-10-10
Allow Frigate port to be switched to authenticated port

### LINKED ISSUE #2860: Frigate uses unauthenticated port for web portal and k8s svc
See https://github.com/truenas/charts/issues/2803 for info

https://github.com/truenas/charts/blob/master/library/ix-dev/community/frigate/templates/_service.tpl#L14

adyanth:
Fixes #2860


stavros-k:
Hello, this will not migrate correctly to the new apps systems.
And because migrations are already in place. there are no plans to allow features added on charts.

PR #2896, opened 2024-10-27
nextcloud: update postgres to 15.8

### LINKED ISSUE #2895: Update postgres version for nextcloud
I've been working on migrating everything off of TrueCharts and when manually updating Nextcloud to version 30 so that I can dump the database and restore with the TrueNAS chart it seems my database was updated to Postgres 1.15, thus preventing me from restoring my database dump on the TrueNAS official chart.

I dug around a bit but I'm not entirely sure how the charts are laid out to figure out where I'd find this to make a PR, I'll continue digging and if I find it I'll make a PR.

Renari:
resolves #2895


stavros-k:
Sorry but doing that will break all existing installations.
You cant just bump 2 major versions. You need to handle upgrades.
Also no big changes will happen on the charts catalog, as this will also break migrations to the new apps catalog.

You can follow this https://github.com/truenas/apps/issues/342

PR #3023, opened 2024-12-28
Frigate: Add new authenticated UI / API port to app

janishutz:
I have been setting up Frigate and noticed that the TrueNAS Scale app didn't yet have the authenticated UI port forwarded, so I went on to add that. It probably isn't correct just yet, just tell me what is incorrect or fix it yourself. Thanks for y'all's great work!


stavros-k:
Hello, auth port exists in the new apps.
https://github.com/truenas/apps/blob/f886be723b410ad8c60af75f6f43088f10bc003c/ix-dev/community/frigate/questions.yaml#L120

There are no plans to add it for apps on 24.04 and prior.

Thanks

PR #3074, opened 2025-02-05
Upgraded catalog item(s)

flymin:
This commit upgrades immich catalog item(s).


stavros-k:
As you will notice, tests are failing.
There won't be any more updates for apps that fail tests, because if there are any changes here, migrations will break.

Migrate to Electric Eel if you want more recent Immich app.

Thanks

PR #3110, opened 2025-11-23
Update Redis image tag to use SHA256 digest

rtucker11:
the bitnami/redis team changed from semantic versioning for tags to sha hashes. This is the most recent.


rtucker11:
Not positive if this is rquired or not. I was on an older version, upgrading seems to have resolved.

